PR #2, opened 2017-06-12
Don't import from 'rxjs'

wKoza:
Refer to #1


hansl:
Oh I fixed that in my local build, but forgot to push. LGTM, thanks!

PR #20, opened 2017-07-04
Ng schematics pipe

jschwarty:
The `module` option still needs to be implemented. I didn't see an example of that yet in the `component` schematic so I am not sure how the desired implementation should go (looks like it will probably need a `resolveModulePath` utility function).

@Brocco can you check this out and let me know if I got everything (besides the `module` option) please? 😄

### DIFF
--- a/packages/schematics/angular/pipe/index.ts
+++ b/packages/schematics/angular/pipe/index.ts
@@ -1,0 +1,91 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+// TODO: replace `options: any` with an actual type generated from the schema.
+// tslint:disable:no-any
+import {addDeclarationToModule} from '../utility/ast-utils';
+import {InsertChange} from '../utility/change';
+
+import {
+  Rule,
+  Tree,
+  apply,
+  branchAndMerge,
+  chain,
+  filter,
+  mergeWith,
+  move,
+  noop,
+  template,
+  url,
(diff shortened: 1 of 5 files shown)


packages/schematics/angular/pipe/index.ts
```diff
@@ -0,0 +1,111 @@
...
+} from '@angular-devkit/schematics';
+import * as stringUtils from '../strings';
+
+import * as ts from 'typescript';
+import 'rxjs/add/operator/merge';
+
+
+function addDeclarationToNgModule(options: any): Rule {
+  return (host: Tree) => {
+    if (options['skip-import']) {
```
Splaktar: This should probably be consistent with https://github.com/angular/devkit/blob/master/packages/schematics/angular/component/index.ts#L33? I'm not sure which of the two should be preferred.

packages/schematics/angular/pipe/schema.json
```diff
@@ -0,0 +1,47 @@
...
+      "type": "boolean",
+      "default": false,
+      "description": "Flag to indicate if a dir is created."
+    },
+    "spec": {
+      "type": "boolean",
+      "default": true,
+      "description": "Specifies if a spec file is generated."
+    },
+    "skip-import": {
```
Splaktar: Should be `"skipImport"`?

jschwarty:
Thanks for the initial feedback @Splaktar! Got those code format changes made and changed skip-import to camel case.

Brocco:
This looks good, sans the module import code which I'm refactoring for the `angular-app` schematic. I should have that pushed in a few hours which will allow you to update this PR.

Brocco:
@jschwarty PR #29 contains the updates for module resolutions.

PR #24, opened 2017-07-07
Schematic updates

Brocco:


Brocco:
Commit message fixes and rebased into a single commit.

PR #26, opened 2017-07-10
Ng schematics service

jschwarty:
Have a `TODO` in there for the module name option as per discussion with @Brocco. Also, the `package-lock.json` file is in here. Not sure what the plan is to handle that as it changes from dev to dev. Let me know if I need to do anything on that.

### DIFF
--- a/packages/schematics/angular/service/index.ts
+++ b/packages/schematics/angular/service/index.ts
@@ -1,0 +1,82 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+// TODO: replace `options: any` with an actual type generated from the schema.
+// tslint:disable:no-any
+import {
+  Rule,
+  Tree,
+  apply,
+  branchAndMerge,
+  chain,
+  filter,
+  mergeWith,
+  move,
+  noop,
+  template,
+  url,
+} from '@angular-devkit/schematics';
+import 'rxjs/add/operator/merge';
+import * as ts from 'typescript';
+import * as stringUtils from '../strings';
+import {addProviderToModule} from '../utility/ast-utils';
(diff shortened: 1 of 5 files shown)


Brocco:
@jschwarty PR #29 contains the updates for module resolutions.

packages/schematics/angular/service/index.ts
```diff
@@ -0,0 +1,86 @@
...
+import {buildRelativePath} from '../utility/find-module';
+
+function addProviderToNgModule(options: any): Rule {
+  return (host: Tree) => {
+    if (!options.module) {
+      return host;
+    }
+
+    let modulePath;
+    if (options.module) {
```
Brocco: This check is redundant (see line 33)
Once that's removed `modulePath` can be a const instead of being `let` and set on line 42.

package-lock.json
```diff
@@ -12,10 +12,14 @@
     "@ngtools/logger": {
```
Brocco: There should be no changes to `package-lock.json` in this PR either

hansl: That's a bad merge. just reset it.

packages/schematics/angular/service/index.ts
jschwarty: Something like this?
```
    if (!options.module) {
      return host;
    }

    const modulePath = options.module;
    if (!host.exists(options.module)) {
      throw new Error(`Module specified (${options.module}) does not exist.`);
    }
```

packages/schematics/angular/service/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  branchAndMerge,
+  chain,
+  filter,
+  mergeWith,
+  move,
+  noop,
+  template,
+  url,
+} from '@angular-devkit/schematics';
+import * as stringUtils from '../strings';
```
hansl: Please merge all import groups together (this is different than the CLI repo).

jschwarty: So just no empty line here between `import` lines, correct?

PR #30, opened 2017-07-13
feat(@schematics/angular): Add guard schematic

Brocco:


packages/schematics/angular/guard/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  branchAndMerge,
+  chain,
+  filter,
+  mergeWith,
+  move,
+  noop,
+  template,
+  url,
+} from '@angular-devkit/schematics';
+import * as stringUtils from '../strings';
```
hansl: Could you merge those groups of imports?

PR #43, opened 2017-07-19
fix(@schematics/angular): Pin version of zone.js and resolve name iss…

Brocco:
…ues in module schematic"


packages/schematics/angular/angular-app/files/package.json
```diff
@@ -22,7 +22,7 @@
     "@angular/router": "^4.0.0",
     "core-js": "^2.4.1",
     "rxjs": "^5.1.0",
-    "zone.js": "^0.8.4"
+    "zone.js": "0.8.12"
```
filipesilva: Unpin to `^0.8.14`, it was released yesterday.

PR #50, opened 2017-07-25
Schematics cleanup

Brocco:


packages/schematics/angular/angular-app/index.ts
```diff
@@ -103,6 +109,8 @@ export default function (options: any): Rule {
     addBootstrapToNgModule(options.directory),
     mergeWith(
       apply(url('./other-files'), [
+        options.minimal || options.inlineTempalte ?
```
filipesilva: `inlineTempalte ` -> `inlineTemplate `

packages/schematics/angular/angular-app/index.ts
```diff
@@ -103,6 +109,8 @@ export default function (options: any): Rule {
...
+          filter(path => !path.endsWith('.html')) : noop(),
```
filipesilva: Don't you need to do something similar for `options.inlineStyle`?

Brocco: No, this is only because we replace the template with the default app one from the `other-files` dir.

packages/schematics/angular/angular-app/index.ts
Brocco: Good catch, fixed.

PR #56, opened 2017-07-29
docs(@schematics/angular): update url in docs

elhadiahmed:
update the reference cli url in documentation

resolve issue #55

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,15 +1,19 @@
 version: 2
 
-defaults: &defaults
+_defaults: &defaults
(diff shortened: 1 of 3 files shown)


hansl:
That's not the right reference. The CLI can be found here: https://github.com/angular/devkit/blob/master/packages/angular_devkit/schematics/bin/schematics.ts

elhadiahmed:
@hansl I updated the reference cli url in documentation to point to the schematics.ts file, I thought that the whole project is the one that needs to be referenced, sorry for the confusion.

PR #61, opened 2017-08-02
docs(@schematics/angular): update url in docs

elhadiahmed:
update the reference cli url in documentation to point to the schematics.ts file

resolve issue #55


elhadiahmed:
@hansl looks good ? I was wondering what should I do regarding the failing tests ?

hansl:
It's fine. Seems like the tests are flaky. I'll bypass.

PR #72, opened 2017-08-04
build: remove marked-terminal

filipesilva:
We're not using it and it was showing a missing peerDep for `marked`.


hansl:
Note: I will want to use it (or something similar) at some point when implementing the CLI. For now it's not useful though so removing it.

PR #73, opened 2017-08-04
fix(@angular-devkit/build-optimizer): make cli logging consistent

### LINKED ISSUE #68: Build Optimizer: Inconsistent logging
In [cli.ts](https://github.com/angular/devkit/blob/630d3f5734a45897d302b68f76ae54b3ad3f6cdb/packages/angular_devkit/build_optimizer/src/build-optimizer/cli.ts#L41-L46) a message is only logged when nothing is emitted, a message should be shown to the user when something is emitted.

filipesilva:
Fix #68


hansl:
At some point in the future (like soon), CLI will depend on the devkit core, which includes our logger classes and utilities. We should probably move the build-optimizer to use that, what do you think?

filipesilva:
Definitely.

Issue #75, opened 2017-08-05
Incorrect Implementations of Schematic interface


elhadiahmed:
I get the below lint error in the SchematicImpl class in [schematic.ts](https://github.com/angular/devkit/blob/master/packages/angular_devkit/schematics/src/engine/schematic.ts)
```Class 'SchematicImpl<CollectionT, SchematicT>' incorrectly implements interface 'Schematic<CollectionT, SchematicT>'.
  Types of property 'call' are incompatible.
    Type '<OptionT extends object>(options: OptionT, host: Observable<Tree>) => Observable<Tree>' is not assignable to type '<T>(options: T, host: Observable<Tree>) => Observable<Tree>'.
      Types of parameters 'options' and 'options' are incompatible.
        Type 'T' is not assignable to type 'object'.```

hansl:
The right syntax for automatically closing issues in github is `Fixes #123`.

This issue is fixed by https://github.com/angular/devkit/pull/76

elhadiahmed:
Noted.

PR #76, opened 2017-08-05
fix(@angular-devkit/devkit): fix incorrect implementation of schematic interface

elhadiahmed:
fix the SchematicImpl class incorrectly implementing the Schematic interface

resolve issue #75

### DIFF
--- a/packages/angular_devkit/schematics/src/engine/interface.ts
+++ b/packages/angular_devkit/schematics/src/engine/interface.ts
@@ -101,7 +101,7 @@ export interface Schematic<CollectionMetadataT extends object, SchematicMetadata
   readonly description: SchematicDescription<CollectionMetadataT, SchematicMetadataT>;
   readonly collection: Collection<CollectionMetadataT, SchematicMetadataT>;
 
-  call<T>(options: T, host: Observable<Tree>): Observable<Tree>;
+  call<OptionT extends object>(options: OptionT, host: Observable<Tree>): Observable<Tree>;
 }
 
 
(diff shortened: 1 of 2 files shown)


elhadiahmed:
yes, sorry about that this one should cover it I believe unless I am missing something else, I am still new to the open source contributions and I really appreciate your patience 😄

elhadiahmed:
@hansl I tried to ammend my commit message with the correct scope (same name of inside the package.json) but I still get the same error on the ci tool

isn't the packageJsonPaths used in the validate-commits.ts file is reading all package.json files inside the repo and adding it to the packages array ?

```
b477fb5 fix(@angular-devkit/devkit): fix the dependencies of the corrected schematic interface
  The following commit has a scope that does not exist:
```

hansl:
You need to squash your commits so there's only 1 commit left with that line. We do not allow merge commits on master. Here's a quick guide on how to squash: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

In the future, you can change the top commit by using `git commit --amend`, and force pushing to your original branch.

Thanks!

hansl:
PS: right now you have 5 commits in this PR, so you would need to squahs the last 4 into the first one, and rewrite the commit message for that commit. The error is coming from your merge commit which we don't allow.

elhadiahmed:
@hansl I am a bit puzzled regarding the scope error, isn't ```@angular-devkit/devkit``` is the correct scope name ?

```Found 1 commits...
  The following commit has a scope that does not exist:
    929f613 fix(@angular-devkit/devkit): fix incorrect implementation of schematic interface
1 commits were found invalid...
A fatal error happened. See details above.
```

elhadiahmed:
@hansl the integration tests are timing out, is there something I can do about it ? or shall you rebuild ?

hansl:
It's okay _for this PR_ to ignore :)

elhadiahmed:
Great, thank you so much.

PR #77, opened 2017-08-05
fix(@angular-devkit/build-optimizer): don't optimizer service-worker

### LINKED ISSUE #78: Build optimizer: removing BrowserAnimationsModule
"@angular/cli": "1.3.0-rc.5"
"typescript": "2.4.2" ( and 2.3.4)

**Workaround**: `ng build --preserve-symlinks` allows to work with `npm link`ed modules,
and remove tsconfig path entries

**Issue**:
BrowserAnimationsModule is completely removed from `main.js` bundle
which leads to issue in `_createProviderInstance`
here: https://github.com/angular/angular/blob/39b92f7e546b8bdecc0e08c915d36717358c122b/packages/core/src/view/ng_module.ts#L110

because `ctor` will be undefined

How to reproduce
1. `ng new test-bob`
2. in `app.module.ts` add
`import { BrowserAnimationsModule } from '@angular/platform-browser/animations';`
and 
```js
  imports: [
    BrowserModule,
    BrowserAnimationsModule
  ],
```
3. in `tsconfig.app.json` add line to `paths`
```json
"types": [],
    "paths": {
      "@angular/*": ["../node_modules/@angular/*"]
    }
```
Why add this kind of mapping?
It is needed for successful build when you work with linked angular modules

filipesilva:
Fix #78
Fix https://github.com/angular/angular-cli/issues/7280

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
+++ b/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
@@ -16,7 +16,22 @@ import { getScrubFileTransformer } from '../transforms/scrub-file';
 const hasDecorators = /decorators/;
 const hasCtorParameters = /ctorParameters/;
 const hasTsHelpers = /var (__extends|__decorate|__metadata|__param) = /;
-const isAngularPackage = /(\\|\/)node_modules(\\|\/)@angular(\\|\/)/;
+const isAngularModuleFile = /\.es5\.js$/;
+const whitelistedAngularModules = [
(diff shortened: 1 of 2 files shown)


hansl:
Wouldn't it be better to work with a whitelist rather than a blacklist here? There will be more packages added to `@angular` in the future and we might want to opt them in rather than out (so current versions are future proof).

filipesilva:
@hansl added whitelist and check to make sure it only optimizes `.es5.js` files (see #78).

zackarychapple:
Is there a targeted version of the CLI that this will be landing in?

filipesilva:
Next 1.3.x release.

zackarychapple:
Awesome, thanks!

PR #80, opened 2017-08-08
fix(@schematics/angular): Correct module import logic

Brocco:


packages/schematics/angular/module/index.ts
```diff
@@ -46,7 +46,7 @@ function addDeclarationToNgModule(options: any): Rule {
                      + stringUtils.dasherize(options.name)
                      + '.module';
     const relativePath = buildRelativePath(modulePath, importModulePath);
-    const changes = addDeclarationToModule(source, modulePath,
+    const changes = addImportToModule(source, modulePath,
                                            stringUtils.classify(`${options.name}Module`),
```
hansl: alignment, please fix.

Brocco: done

PR #84, opened 2017-08-09
fix(@schematics/angular): Updates for alignment

Brocco:


packages/schematics/angular/directive/index.ts
```diff
@@ -90,9 +90,9 @@ function addDeclarationToNgModule(options: any): Rule {
 
 
 function buildSelector(options: any) {
-  let selector = stringUtils.dasherize(options.name);
+  let selector = stringUtils.classify(options.name);
```
hansl: This sounds really wrong to me.

PR #85, opened 2017-08-09
fix(@angular-devkit/build-optimizer): prefix crt function calls

kara:
We currently have a special purify rule to prefix top-level function calls from generated code, but it doesn't yet include the `createRendererType2` call (aka `ɵcrt`). Retaining this pulls in a bunch of other symbols that we don't need (e.g. styles from Material components).


filipesilva:
@kara can you add a sample line for it in the unit test?
https://github.com/angular/devkit/blob/master/packages/angular_devkit/build_optimizer/src/purify/purify_spec.ts#L144-L154

kara:
Fair point; still in prototype mode :) I'll add it to the tests.

PR #94, opened 2017-08-10
fix(@angular-devkit/schematics): Handle unset paths when normailzing

Brocco:


packages/angular_devkit/schematics/src/utility/path.ts
```diff
@@ -45,6 +45,11 @@ export function relativePath(from: SchematicPath, to: SchematicPath): SchematicP
 
 export function normalizePath(path: string): SchematicPath {
   let p = path;
+
+  if (p === undefined || p === null) {
```
hansl: oh no. We do not want undefined paths.

PR #95, opened 2017-08-11
Various Fixes

hansl:


packages/schematics/angular/module/index.ts
```diff
@@ -64,23 +64,25 @@ function addDeclarationToNgModule(options: any): Rule {
 }
 
 export default function (options: any): Rule {
-  options.module = normalizePath(options.module);
+  return (host: Tree, context: SchematicContext) => {
+    options.module = findModuleFromOptions(host, options);
```
Brocco: This should only be called if `options.module` has a value... modules aren't automatically found.

packages/schematics/angular/service/index.ts
```diff
@@ -63,23 +63,25 @@ function addProviderToNgModule(options: any): Rule {
 }
 
 export default function (options: any): Rule {
-  options.module = normalizePath(options.module);
+  return (host: Tree, context: SchematicContext) => {
+    options.module = findModuleFromOptions(host, options);
```
Brocco: This should only be called if `options.module` has a value...services are not automatically provided.

packages/schematics/angular/utility/find-module.ts
```diff
@@ -5,12 +5,46 @@
...
+  sourceDir?: string;
+  path?: string;
+}
+
+
+/**
+ * Find the module refered by a set of options passed to the schematics.
+ */
+export function findModuleFromOptions(host: Tree, options: ModuleOptions): SchematicPath {
+  if (!options.module) {
```
Brocco: This must also handle the case if `options.skipImport` is set to true.

PR #103, opened 2017-08-14
Test improvements

filipesilva:
See individual commits for details.

### DIFF
--- a/scripts/test.ts
+++ b/scripts/test.ts
@@ -9,6 +9,7 @@ import * as glob from 'glob';
 import * as Istanbul from 'istanbul';
 import 'jasmine';
 import {SpecReporter as JasmineSpecReporter } from 'jasmine-spec-reporter';
+import { ParsedArgs } from 'minimist';
 import { join, relative } from 'path';
 import { Position, SourceMapConsumer, SourceMapGenerator } from 'source-map';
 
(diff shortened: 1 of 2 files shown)


scripts/test.ts
```diff
@@ -201,12 +202,13 @@ glob.sync('packages/**/*.spec.ts')
     console.error(`Invalid spec file name: ${path}. You're using the old convention.`);
   });
 
-// Run the tests.
-const allTests =
-  glob.sync('packages/**/*_spec.ts')
-    .map(p => relative(projectBaseDir, p))
-    .filter(p => !/schematics_cli\/schematics\//.test(p));
+export default function (args: ParsedArgs) {
+  const regex = `packages/${args.glob ? `**/${args.glob}/` : ''}**/*_spec.ts`;
```
hansl: I think this could be easier to follow:

```typescript
let regex = ‘packages/**/*_spec.ts’;
if (args.glob) {
  regex = `packages/**/${args.glob}/**/*_spec.ts`;
}
```

What do you think?

scripts/test.ts
```diff
@@ -196,7 +196,7 @@ runner.onComplete((success: boolean) => {
 
 
 glob.sync('packages/**/*.spec.ts')
-  .filter(p => !/\/schematics\/.*\/files\//.test(p))
+  .filter(p => !/\/schematics\/.*\/(other-)?files\//.test(p))
```
hansl: Non-blocking comment:

Yeah I’ve been thinking about it, and I don’t think this is the right fix (we will have other things than « other-files » as a directory name).

I’m starting to think the right fix would be to move files into a subdirectory of files and other-files into a different subdir under files. But that’s more involved in general, so I’ll approve this. It’s a convention.

filipesilva: Yeah I agree. @Brocco would that be reasonable?

scripts/test.ts
filipesilva: I agree. Done.

PR #110, opened 2017-08-20
build(@angular/cli): ignore vim and emacs files

redbmk:
Change the default `build.ignored` option to ignore
files used by vim and emacs that are not needed by
angular-cli

This depends on changes from angular/angular-cli#7448

### DIFF
--- a/packages/schematics/angular/application/files/__dot__angular-cli.json
+++ b/packages/schematics/angular/application/files/__dot__angular-cli.json
@@ -51,6 +51,9 @@
     }
   },
   "defaults": {
+    "build": {
+      "ignored": ["**/*~", "**/.#*", "**/.*.sw[pon]"]
+    },
     "styleExt": "<%= style %>",<% if (!minimal) { %>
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/application/files/__dot__angular-cli.json
```diff
@@ -51,6 +51,9 @@
     }
   },
   "defaults": {
+    "build": {
+      "ignored": ["**/*~", "**/.#*", "**/.*.sw[pon]"]
```
hansl: Actually, nevermind my last comment. We don't have an option in the configuration for `ignored`. Where did you get that field from?

redbmk:
@hansl Sorry, this PR was a bit premature - it's really part of angular/angular-cli#7448, it's just that they're in separate repositories so I wasn't sure whether to wait for that one to get approved or just do them both at the same time. This PR relies on the other one, but it looks like there were some changes that made that one unnecessary (and therefore this one as well).

PR #114, opened 2017-08-23
Add ability to define aliases by extending existing schematics.

Brocco:


packages/angular_devkit/schematics/src/engine/engine.ts
```diff
@@ -85,6 +85,19 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
     if (!description) {
       throw new UnknownSchematicException(name, collection);
     }
+    if (!!description.extends) {
+      const parts = description.extends.split(':');
+      const collectionName = parts[0];
+      const schematicName = parts[1] || parts[0];
+      if (schematicName) {
```
hansl: This condition will always be true.

packages/angular_devkit/schematics/src/engine/engine.ts
```diff
@@ -85,6 +85,19 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
     if (!description) {
       throw new UnknownSchematicException(name, collection);
     }
+    if (!!description.extends) {
+      const parts = description.extends.split(':');
+      const collectionName = parts[0];
```
hansl: You don't have a fallback if the `extends` does not contain a `:`.

packages/angular_devkit/schematics/src/engine/engine.ts
```diff
@@ -85,6 +85,19 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
     if (!description) {
       throw new UnknownSchematicException(name, collection);
     }
+    if (!!description.extends) {
+      const parts = description.extends.split(':');
```
hansl: Could we add a limit to this `split`? Basically, if there are multiple `:` we should allow `:` to be part of the schematics name. WDYT?

packages/angular_devkit/schematics/tools/description.ts
```diff
@@ -25,6 +25,7 @@ export interface FileSystemSchematicJsonDescription {
   readonly factory: string;
   readonly description: string;
   readonly schema?: string;
+  readonly extends?: string;
```
hansl: This is not specific to file system description, and it's already included in the description type anyway. Remove it from here.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -58,7 +58,22 @@ export abstract class FileSystemEngineHostBase implements
...
+        schematics.push(collectionName || schematicName);
+      } else {
+        schematics.push(key);
+      }
+    }
+    return schematics.reduce((acc, curr) => {
+      if (acc.indexOf(curr) === -1) {
+        acc.push(curr);
+      }
+      return acc;
```
hansl: lint.

packages/angular_devkit/schematics/src/engine/engine.ts
Brocco: I think that's addressing an edge case, but since it'd be simple to do, sure.

packages/angular_devkit/schematics/tools/description.ts
Brocco: Both need to be there... unless the two types are not correctly deriving from the other.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -108,6 +125,14 @@ export abstract class FileSystemEngineHostBase implements
       return null;
     }
 
+    if (partialDesc.extends) {
+      return {
```
hansl: We probably want to fallback to the rest of the function, so people can extend and change some argument like `factory`.

PR #119, opened 2017-08-24
Add --open parameter to start script

winksaville:
This causes executing the start script to open the angular
app automatically.


hansl:
Hi @winksaville,

Unfortunately, after discussing this a bit offline, we found that this isn't something we want to implement.

It's trivial for you to change once you created a project and not something most people want. I'm going to close this PR. Thanks.

winksaville:
Thanks for considering it.

PR #128, opened 2017-08-30
WIP: Refactor Proof Of Concept

hansl:


package.json
```diff
@@ -63,6 +64,7 @@
     "npm": "^5.2.0",
     "npm-run": "^4.1.2",
     "npm-run-all": "^4.0.0",
+    "parse5": "^3.0.2",
```
lacolaco: does arborist use parse5 rather than domino? (about consistency to angular/angular dependencies)

hansl: Parse5 is for parsing the HTML. Domino is a DOM implementation, which Universal uses.

Angular parses HTML using their own parser, which has some specific details that we might want.

PR #129, opened 2017-09-04
fix(@schematics/angular): fix missing link attribute for PWA

markgoho:
This change fixes a [security/performance issue](https://developers.google.com/web/tools/lighthouse/audits/noopener) in links that open in new windows, and by the same token, increases the lighthouse score for PWAs. Assuming we (eventually) want a CLI-generated Angular project to be 100% PWA, this is a necessary change.

image

### DIFF
--- a/packages/schematics/angular/application/other-files/app.component.html
+++ b/packages/schematics/angular/application/other-files/app.component.html
@@ -8,13 +8,13 @@
 <h2>Here are some links to help you start: </h2>
 <ul>
   <li>
-    <h2><a target="_blank" href="https://angular.io/tutorial">Tour of Heroes</a></h2>
+    <h2><a target="_blank" rel="noopener" href="https://angular.io/tutorial">Tour of Heroes</a></h2>
   </li>
   <li>
-    <h2><a target="_blank" href="https://github.com/angular/angular-cli/wiki">CLI Documentation</a></h2>
+    <h2><a target="_blank" rel="noopener" href="https://github.com/angular/angular-cli/wiki">CLI Documentation</a></h2>
   </li>
   <li>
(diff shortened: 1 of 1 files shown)


filipesilva:
Heya, can you change the commit message to `fix(@schematics/angular): fix missing link attribute for PWA` please? That's what's failing the CI. Other than that LGTM.

markgoho:
Updated title, sorry abou that @filipesilva -- thanks for noticing this PR so quickly!

filipesilva:
@markgoho what I actually meant is the commit itself though, not the PR title. You can rename it by doing `git commit --amend`.

markgoho:
Okay I did `git commit --amend -m "fix(@schematics/angular): fix missing link attribute for PWA"` and then pushed. Let me know if that made it right.

filipesilva:
You got the right message, but now you have 3 commits, some of which have invalid messages.

A good way of squashing them all into one is to use `git rebase master -i`. On the first commit replace the `pick` word with `reword` and on the other three replace `pick` with `fixup`. This will give you a single final commit.

markgoho:
Okay, I thought I rebased it properly, but it looks like CI is failing again :(

markgoho:
Here's what my text editor looks like just before I hit save
image
Maybe `fixup` isn't the right keyword?

filipesilva:
Something has gone wrong, and I don't really know what it was :(

In my editor, rebase -i looks like this:
image

So I think fixup is right... but something obviously did not go according to plan. It seems a bunch of merge commits found their way in?

I think at this point it's easier to make a new PR since your changes are limited to a single file. I'm sorry this process is being so dragged out, and thanks for sticking through it!

markgoho:
Thanks for all of your effort too! I'll close this and re-open with the correct commit message :)

PR #132, opened 2017-09-06
Fix the filter for ngo's purify webpack plugin

danbucholtz:
This PR fixes the filter in the NGO purify plugin to include files that end in `.js` instead of `.bundle.js`.

I discussed this PR with @filipesilva on slack.

Thanks,
Dan


danbucholtz:
Fixing commit msg

PR #139, opened 2017-09-07
feat(@schematics/angular): remove codelyzer rules

wKoza:
It's a copy/paste of this [PR](https://github.com/angular/angular-cli/pull/6651) :

Few codelyzer rules report warnings which are also compile-time errors caught by the
Angular compiler and reported by the language service.

Since codelyzer doesn't aim to provide type checking and deep analysis
of the entire project, it'll be better if these two rules are removed.

Btw, soon more rules will follow mgechev/codelyzer#314.

cc/ @mgechev


mgechev:
Yes, removing these two rules makes sense. They are related to compile-time errors handled by the compiler.

PR #140, opened 2017-09-07
chore: update to latest bazel rules

alexeagle:


alexeagle:
don't merge yet, we should be able to include a windows fix in here

alexeagle:
okay you can merge when you like, @hansl

PR #147, opened 2017-09-12
refactor(@angular-devkit/build-optimizer): use single pass to wrap enums

clydin:
This converts the enum wrapping transformer to use a single pass of the AST and replaces enum declarations inline.  It also attempts to reuse existing structure where possible.  The enum member declaration _finder_ functions were not modified in this PR.

//cc @filipesilva


filipesilva:
LGTM, thanks for improving this transformer.

PR #149, opened 2017-09-14
App root schematics

hansl:


packages/schematics/angular/service/schema.json
```diff
@@ -11,6 +11,10 @@
       "type": "string",
       "default": "app"
     },
+    "appRoot": {
+      "type": "string",
+      "default": "app"
```
filipesilva: We shouldn't have a default to ensure backwards compatibility with CLI versions that aren't sending the appRoot options.

hansl: done.

PR #153, opened 2017-09-15
fix(@schematics/angular): restore router-outlet

cyrilletuzi:
Add missing router-outlet in inline AppComponent.

Fixes angular/angular-cli#7701


filipesilva:
@Brocco can you review?

cyrilletuzi:
Added the condition to put `router-outlet` only if routing was asked.

cyrilletuzi:
Is it possible to have a patch release for this ? I'll do an Angular course on Monday, and it would be nicer without this issue.

PR #154, opened 2017-09-18
fix(@schematics/angular): fix inconsistency in service schema

RicardoVaranda:
Closes this issue:

https://github.com/angular/angular-cli/issues/7679


filipesilva:
@Brocco can you have a look?

PR #157, opened 2017-09-20
fix(@angular-devkit/build-optimizer): fix purify sourcemaps

filipesilva:
/cc @kara @IgorMinar

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/purify/purify.ts
+++ b/packages/angular_devkit/build_optimizer/src/purify/purify.ts
@@ -5,38 +5,89 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import { RawSource, ReplaceSource } from 'webpack-sources';
+
+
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_optimizer/src/purify/purify.ts
```diff
@@ -5,38 +5,99 @@
...
+  /* Prefix module statements */
+  content.replace(
+    /new \w*__WEBPACK_IMPORTED_MODULE_\d+__angular_core__\["\w+" \/\* NgModuleFactory \*\/\]/mg,
+    (match, offset) => {
+      const newContent = `/*@__PURE__*/${match}`;
+      replacements.push({
+        start: offset,
+        end: offset + match.length - 1,
+        content: newContent,
+      });
```
kara: Nit: Lines 48 - 54 are pretty similar to 63 - 69 and 78-84. Could we re-use any of the code in a shared function?

filipesilva: They are indeed, fixed.

PR #164, opened 2017-09-22
fix(@angular-devkit/schematics): fix extends to properly load external collections

vsavkin:
Fixes loading external collections with extends

### DIFF
--- a/packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
+++ b/packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
@@ -136,7 +136,7 @@ export abstract class FileSystemEngineHostBase implements
     }
 
     const collectionPath = dirname(collection.path);
-    let partialDesc: Partial<FileSystemSchematicDesc> | null = collection.schematics[name];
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -136,58 +136,57 @@ export abstract class FileSystemEngineHostBase implements
...
-    // Use any on this ref as we don't have the OptionT here, but we don't need it (we only need
-    // the path).
-    if (!partialDesc.factory) {
-      throw new SchematicMissingFactoryException(name);
-    }
-    const resolvedRef = this._resolveReferenceString(partialDesc.factory, collectionPath);
-    if (!resolvedRef) {
-      throw new FactoryCannotBeResolvedException(name);
-    }
+    } else {
```
hansl: Since all paths in the previous `if` returns, removing this `else` would simplify the diff greatly and also remove one level of indentation.

vsavkin:
Fixed

PR #165, opened 2017-09-25
feat(@schematics/angular): add interceptor schematic

delasteve:


delasteve:
Will fix lint issues in the morning.

filipesilva:
@Brocco can you take a look?

alexeagle:
Sorry @delasteve  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #166, opened 2017-09-25
feat(@schematics/angular): update tslint and codelyzer to latest

clydin:


hansl:
This broke the CLI end-to-end tests because fixing errors now returns a zero status (it didn't before) when there's no more linting issue (test `tests/lint/lint-with-type-check-fail`).

PR #167, opened 2017-09-25
fix(@schematics/angular): Add skip git option to prevent creating git…

Brocco:
…ignore file

fixes angular/angular-cli#7686

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -94,6 +94,7 @@ export default function (options: ApplicationOptions): Rule {
       mergeWith(
         apply(url('./files'), [
           options.minimal ? filter(minimalPathFilter) : noop(),
+          options.skipGit ? filter(path => !path.endsWith('/__dot__gitignore')) : noop(),
           template({
             utils: stringUtils,
             'dot': '.',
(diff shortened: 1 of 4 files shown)


packages/schematics/angular/application/index.ts
```diff
@@ -94,6 +94,7 @@ export default function (options: ApplicationOptions): Rule {
       mergeWith(
         apply(url('./files'), [
           options.minimal ? filter(minimalPathFilter) : noop(),
+          options.skipGit ? filter(path => !path.endsWith('gitignore')) : noop(),
```
hansl: Could you check the whole path?  `!path.endsWith('/__dot__gitignore')`?

packages/schematics/angular/application/schema.d.ts
```diff
@@ -25,11 +25,11 @@ export interface Schema {
     /**
      * Specifies the view encapsulation strategy.
      */
-    viewEncapsulation?: ('Emulated' | 'Native' | 'None');
+    viewEncapsulation?: ("Emulated" | "Native" | "None");
```
hansl: Unfortunately you have to do a `devkit-admin lint --fix` run after generating the `.d.ts`. This won't pass linting.

packages/schematics/angular/application/index.ts
Brocco: fixed this (plus linting issue in generated schema file)

PR #169, opened 2017-09-26
Build optimizer es2015

filipesilva:
This reverts https://github.com/angular/devkit/pull/168 because this es2015 support needed more than that.

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/transforms/scrub-file_spec.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/scrub-file_spec.ts
@@ -173,6 +173,20 @@ describe('scrub-file', () => {
       expect(oneLine`${transform(input)}`).toEqual(oneLine`${output}`);
     });
 
+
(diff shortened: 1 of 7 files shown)


packages/angular_devkit/build_optimizer/src/transforms/scrub-file_spec.ts
```diff
@@ -173,6 +173,20 @@ describe('scrub-file', () => {
       expect(oneLine`${transform(input)}`).toEqual(oneLine`${output}`);
     });
 
+
+    fit('removes top-level Angular constructor parameters in es2015', () => {
```
devoto13: `fit`?

filipesilva: Good catch 👍

filipesilva: Fixed in https://github.com/angular/devkit/pull/170

PR #173, opened 2017-09-27
feat(@angular-devkit/schematics): go further in the caller list

hansl:
So we try more local options for finding the package.json. Also special case
package.json as its not normally resolved by resolve itself.

### DIFF
--- a/packages/angular_devkit/core/node/resolve.ts
+++ b/packages/angular_devkit/core/node/resolve.ts
@@ -10,9 +10,11 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { isFile } from './fs';
 
-
+/**
+ * Exception thrown when a module could not be resolved.
+ */
 export class ModuleNotFoundException extends BaseException {
-  public code: string;
+  public readonly code: string;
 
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/core/node/resolve.ts
```diff
@@ -64,20 +70,57 @@ function _getGlobalNodeModules() {
...
+ * @returns {string} Path to the index to include, or if `resolvePackageJson` option was
+ *                   passed, a path to that file.
+ * @throws {ModuleNotFoundException} If no module with that name was found anywhere.
+ */
 export function resolve(x: string, options: ResolveOptions = {}): string {
   const readFileSync = fs.readFileSync;
 
   const extensions: string[] = options.extensions || Object.keys(require.extensions);
-  const basePath = options.basedir || path.dirname(_caller());
+  const basePath = options.basedir || path.dirname(_caller()[0]);
```
Brocco: Instead of having this arbitrary first element access, the basedir option could be made to be not optional.

hansl: done

PR #176, opened 2017-09-28
refactor: add lint checks for fdescribe and fit tests

delasteve:
Per discussion with @filipesilva


hansl:
Hey @delasteve, instead of adding another dependency, could you just add the rule to the `rules` directory? The code is simple (https://github.com/Sergiioo/tslint-defocus/blob/master/src/defocusRule.ts).

delasteve:
Done. Let me know if there is any other attribution necessary.

PR #178, opened 2017-09-28
fix(@angular-devkit/schematics): unit test module output es2015

Manduro:
closes angular/angular-cli#6525

Works with code coverage for me!


hansl:
Closing this as this is not something we want. Until the main tsconfig moves to es2015 by default, the spec tsconfig should stay the same.

PR #181, opened 2017-09-29
Re-add es2015 support and fix es5 detection

filipesilva:
Thanks to @gkalpak for the regex fix 👍

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
+++ b/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
@@ -16,29 +16,29 @@ import { getWrapEnumsTransformer, testWrapEnums } from '../transforms/wrap-enums
 
 
 const whitelistedAngularModules = [
-  /[\\\/]node_modules[\\\/]@angular[\\\/]animations[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]common[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]compiler[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]core[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]forms[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]http[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]platform-browser-dynamic[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]platform-browser[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]platform-webworker-dynamic[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]platform-webworker[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]router[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]upgrade[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]material[\\\/]/,
-  /[\\\/]node_modules[\\\/]@angular[\\\/]cdk[\\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]animations[\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]common[\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]compiler[\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]core[\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]forms[\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]http[\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]platform-browser-dynamic[\\/]/,
+  /[\\/]node_modules[\\/]@angular[\\/]platform-browser[\\/]/,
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
```diff
@@ -16,20 +16,29 @@ import { getWrapEnumsTransformer, testWrapEnums } from '../transforms/wrap-enums
...
+  /[\\\/]node_modules[\\\/]@angular[\\\/]platform-webworker[\\\/]/,
+  /[\\\/]node_modules[\\\/]@angular[\\\/]router[\\\/]/,
+  /[\\\/]node_modules[\\\/]@angular[\\\/]upgrade[\\\/]/,
+  /[\\\/]node_modules[\\\/]@angular[\\\/]material[\\\/]/,
+  /[\\\/]node_modules[\\\/]@angular[\\\/]cdk[\\\/]/,
+];
+
+const es5AngularModules = [
+  // Angular 4 packaging format has .es5.js as the extension.
+  /.es5.js$/, // Angular 4
```
gkalpak: Probably doesn't matter, but (just to be safe) the `.` should be escaped (to avoid matching `foo.noes5.js` :grin:):

```js
/\.es5\.js$/
```

packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
```diff
@@ -16,20 +16,29 @@ import { getWrapEnumsTransformer, testWrapEnums } from '../transforms/wrap-enums
...
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)http(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)platform-browser-dynamic(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)platform-browser(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)platform-webworker-dynamic(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)platform-webworker(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)router(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)upgrade(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)material(\\|\/)/,
-  /(\\|\/)node_modules(\\|\/)@angular(\\|\/)cdk(\\|\/)/,
+  /[\\\/]node_modules[\\\/]@angular[\\\/]animations[\\\/]/,
```
gkalpak: Nit: You don't need to escape `/` inside character groups (`[...]`).

packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
```diff
@@ -16,20 +16,29 @@ import { getWrapEnumsTransformer, testWrapEnums } from '../transforms/wrap-enums
...
+  /[\\\/]node_modules[\\\/]@angular[\\\/]cdk[\\\/]/,
+];
+
+const es5AngularModules = [
+  // Angular 4 packaging format has .es5.js as the extension.
+  /.es5.js$/, // Angular 4
+  // Angular 5 has esm5 folders.
+  /[\\\/]node_modules[\\\/]@angular[\\\/][^\\/]+[\\\/]esm5[\\\/]/,
+  // All Angular versions have UMD with es5.
+  /.umd.js$/,
```
gkalpak: Again, better to escape `.`: `/\.umd\.js$/`

packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
```diff
@@ -70,9 +79,10 @@ export function buildOptimizer(options: BuildOptimizerOptions): TransformJavascr
 
   if (inputFilePath
```
gkalpak: No idea where this path comes form, but is it guaranteed to have a `/` before `node_modules`?
E.g. is there any chance it might start with `node_modules/@angular/...`?

filipesilva: It is guaranteed (in our usage) to be a full, non-relative path.

packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
filipesilva: Fixed

packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
filipesilva: Fixed

packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
filipesilva: Fixed

filipesilva:
Thanks for the reivew @gkalpak 👍

packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
gkalpak: You still need to escape `\` though. So it should be `[\\/]`.
Probably a good idea to add some test too :grin:

Go Windows users! 🥂

filipesilva: That's a really good idea. I should have done that ages ago. Fixed.

PR #197, opened 2017-10-08
Docs

hansl:
This is still early, but it'd be nice to get this in while I'm moving to writing tutorial on Schematics.


.monorepo.json
```diff
@@ -0,0 +1,92 @@
...
+        "url": "[david-badge-url]"
+      },
+      {
+        "title": "![devDependency Status][david-dev-badge]",
+        "url" : "[david-dev-badge-url]"
+      }
+    ], [
+      {
+        "label": "npm",
+        "image": "https://img.shields.io/npm/v/%40angular/cli.svg",
```
filipesilva: A lot of these are still for the CLI.

lib/packages.ts
```diff
@@ -15,6 +15,7 @@ import * as ts from 'typescript';
 
 const glob = require('glob');
 const distRoot = path.join(__dirname, '../dist');
+const { packages: monorepoPackages } = require('../.monorepo.json');
```
filipesilva: 👍

packages/angular_devkit/core/src/utils/template.ts
```diff
@@ -143,6 +155,12 @@ export function templateParser(sourceText: string, fileName: string): TemplateAs
       children.push({ kind: 'escape', expression: escape, start, end } as TemplateAstEscape);
       start = end;
     }
+    if (comment) {
+      end = _positionFor(sourceText, offset + comment.length + 5);
```
filipesilva: What's the meaning of `5` here? Just curious, I see other cases around this one also use it.

scripts/release.ts
```diff
@@ -123,8 +119,8 @@ export default function(args: { _: string[], 'dry-run'?: boolean }, logger: Logg
     case 'patch':
       _upgrade(maybeRelease, logger);
       if (!dryRun) {
-        fs.writeFileSync(path.join(__dirname, '../versions.json'),
-                         JSON.stringify({ versions, hashes }, null, 2) + '\n');
+        fs.writeFileSync(path.join(__dirname, '../.monorepo.json'),
```
filipesilva: Won't this delete all other fields in `monorepo.json` except for `packages`?

scripts/templates/readme.ejs
```diff
@@ -0,0 +1,92 @@
...
+
+  Any changes to README.md directly will result in a failure on CI.
+-->
+
+# Angular DevKit
+### Development tools and libraries specialized for Angular
+
+<%# Badges are created here. %>
+<% for (const section of monorepo.badges) {
+  for (const { githubBadge, label, image, title, url } of section) {
```
filipesilva: `githubBadge` is unused.

packages/angular_devkit/core/src/utils/template.ts
hansl: Number of characters taken from the template itself (the `<%?` and `%>` parts).

.monorepo.json
hansl: done.

scripts/release.ts
hansl: Fixed, you were right. Thanks for catching that!

scripts/templates/readme.ejs
hansl: Removed.

PR #209, opened 2017-10-17
feat(@angular-devkit/build-optimizer): strip Angular decorators in __…

filipesilva:
…decorate call


filipesilva:
Funny, branches starting with `__` don't seem to be updated by github.

PR #210, opened 2017-10-17
feat(@angular-devkit/build-optimizer): strip Angular decorators in __…

filipesilva:
…decorate call

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
@@ -12,6 +12,7 @@ import { collectDeepNodes } from '../helpers/ast-utils';
 export function testScrubFile(content: string) {
   const markers = [
     'decorators',
+    '__decorate',
     'propDecorators',
     'ctorParameters',
   ];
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
```diff
@@ -261,6 +309,36 @@ function pickDecorationNodesToRemove(
...
+  const ngDecoratorCalls = elements.filter((el) => {
+    if (el.expression.kind !== ts.SyntaxKind.Identifier) {
+      return false;
+    }
+    const id = el.expression as ts.Identifier;
+
+    return identifierIsMetadata(id, ngMetadata, checker);
+  });
+
+  return (elements.length > ngDecoratorCalls.length) ? ngDecoratorCalls : [exprStmt];
```
IgorMinar: this logic is hard to follow. maybe you want to say:

```
// if all decorators are metadata decorators then return the whole `Class = __decorate([...])'` statement so that it is removed in entirety
return (elements.length === ngDecoratorCalls.length) ? [exprStmt] : ngDecoratorCalls;
```

filipesilva: Fixed

PR #212, opened 2017-10-17
fix(@schematics/angular): remove the Reflect polyfill

### LINKED ISSUE #206: Refelect Polyfills for Angular no longer necessary
In the file https://github.com/angular/devkit/blob/master/packages/schematics/angular/application/files/__path__/polyfills.ts with the default Polyfills for Angular CLI you have this:

    /** Evergreen browsers require these. **/
    import 'core-js/es6/reflect';
    import 'core-js/es7/reflect';

According to http://kangax.github.io/compat-table/es6/#test-Reflect this seems no longer to be true. By removing these lines, an Angular app can spare some more kB.

hansl:
Only IE11 and less needs it. See http://kangax.github.io/compat-table/es6/#test-Reflect

Fixes #206.

### DIFF
--- a/packages/schematics/angular/application/files/__path__/polyfills.ts
+++ b/packages/schematics/angular/application/files/__path__/polyfills.ts
@@ -37,8 +37,12 @@
 /** IE10 and IE11 requires the following for NgClass support on SVG elements */
 // import 'classlist.js';  // Run `npm install --save classlist.js`.
 
+/** IE10 and IE11 requires the following for the Reflect API. */
+// import 'core-js/es6/reflect';
+
+
 /** Evergreen browsers require these. **/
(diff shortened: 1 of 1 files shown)


trotyl:
The `core-js/es7/reflect` is not `Reflect` in ES7, but the [Metadata Reflection API](https://github.com/rbuckton/reflect-metadata) proposal in no stage, that is [not supported by any browser](http://kangax.github.io/compat-table/esnext/#test-Metadata_reflection_API), but still required for JIT usage.

filipesilva:
That's a really good point... WDYT @hansl?

hansl:
@trotyl That's right, I forgot that we're still using reflect-metadata in JIT (it's not used anymore in AOT). Will update this PR.

PR #217, opened 2017-10-19
fix(@angular-devkit/build-optimizer): fix hoisted class declarations

### LINKED ISSUE #214: build-optimizer over code that's already optimized fails
```
- [ ] feature request
```

### Area
```
- [x] devkit
- [ ] schematics
```

### Repro steps

RxJS 5.5 was recently published. It uses the build-optimizer to produce code that's pre-optimized so it works with applications not running under the CLI. Testing for AIO showed that the build size was increasing using the new features of Rx. Digging in a little further, it appears that when we pull in rxjs and run it through build-optimizer, the existing PURE annotations are removed and not replaced.

You can reproduce this issue using this repo/branch. Look in `main.ts`. Note: actual user-land code will not import from `_esm5` but will instead use mappings to point to it. There's a [pending issue](https://github.com/angular/angular-cli/issues/8096) for CLI to make this change.

https://github.com/jasonaden/repro/tree/rxjs_build_optimizer

### Desired functionality

Should treeshake

### LINKED ISSUE #215: recognize tslib.__decorate as if it was __decorate
```
- [ ] feature request
```

### Area
```
- [x] devkit
- [ ] schematics
```

### Versions
build-optimizer@0.28

### Repro steps
set `importHelpers: true` in tsconfig of an app then build with build optimizer

### Desired functionality
Decorators should be stripped.

This is a follow up on #210 which implements this functionality for global `__decorate` calls.

hansl:
The problem is that hoisted classes dont have the same name, so we need to change
the regex, and to check the end of it for the class name.

Also added a test to verify this works (based on RxJS issue).

Also refactored the complex regex to make a bit more sense.

Fixes #214.
Fixes #215.


packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
```diff
@@ -13,6 +13,7 @@ export function testScrubFile(content: string) {
   const markers = [
     'decorators',
     '__decorate',
+    'tslib.__decorate',
```
IgorMinar: this is not sufficient marker because e.g. typescript 2.5 emits the import of tslib as:

```
import * as tslib_1 from "tslib";
```

packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
```diff
@@ -202,10 +203,22 @@ function isDecorateAssignmentExpression(exprStmt: ts.ExpressionStatement): boole
...
+  let callExprIdent = callExpr.expression as ts.Identifier;
+
   if (callExpr.expression.kind !== ts.SyntaxKind.Identifier) {
-    return false;
+    if (callExpr.expression.kind === ts.SyntaxKind.PropertyAccessExpression) {
+      const propAccess = callExpr.expression as ts.PropertyAccessExpression;
+      const left = propAccess.expression;
+      callExprIdent = propAccess.name;
+
+      if (!(left.kind === ts.SyntaxKind.Identifier && (left as ts.Identifier).text === 'tslib')) {
```
IgorMinar: given `import * as tslib_1 from "tslib";`, we should check if the text matches `tslib(_\d+)?`

or we could trace it back to the import and see if the identifier is an alias for `*` coming from the `"tslib"` module.

packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
```diff
@@ -8,6 +8,9 @@
 import * as ts from 'typescript';
 import { collectDeepNodes } from '../helpers/ast-utils';
 
+const tslibDecorateRe = /\btslib(?:_\d+)?\.__decorate\b/;
+const tslibRe = /\btslib(?:_\d+)?\b/;
```
filipesilva: This name can vary, so it would be safer to look up the import rathar than use a regex. The same is done for Angular symbols in `identifierIsMetadata`+`findAngularMetadata`.

PR #219, opened 2017-10-19
fix(@schematics/angular): missing condition and comma position

cyrilletuzi:
Fixes PR #213 which sorted the options without taking the conditions into consideration, resulting in `class: { spec: false }` for everyone.

Fixes PR #196 comma position.

### DIFF
--- a/packages/schematics/angular/application/files/__dot__angular-cli.json
+++ b/packages/schematics/angular/application/files/__dot__angular-cli.json
@@ -54,15 +54,13 @@
     }
   },
   "defaults": {
-    "styleExt": "<%= style %>",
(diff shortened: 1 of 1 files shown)


cyrilletuzi:
Note that sorting the defaults introduced a condition duplicate, and may result in issues with missing/extra commas if further options are added in the future. It may be better for the CLI to reorder them afterwards.

cyrilletuzi:
@hansl @filipesilva

filipesilva:
@hansl can you take a look?

cyrilletuzi:
There has been several releases of the CLI but this issue is still not fixed, @hansl can you have a look ?

PR #221, opened 2017-10-19
fix(@schematics/angular): build in --prod mode by default

cyrilletuzi:
According to the idea the CLI should be ready to use, if I want to work in dev, I run `npm start`, and if I want to go in production, I run `npm run build`. But currently the build is in dev mode by default, requiring the user to know he/she has to go to modify the npm script accordingly.

@hansl @filipesilva

### DIFF
--- a/packages/schematics/angular/application/files/package.json
+++ b/packages/schematics/angular/application/files/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
-    "build": "ng build",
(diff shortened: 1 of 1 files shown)


filipesilva:
I think this makes sense. @hansl @StephenFluin WDYT?

cyrilletuzi:
@filipesilva Stephen validated this with a reaction, can we merge this ?

hansl:
Yeah we've had this PR open for ages and I still come back to look at it and not close it. I really don't know.

PR #235, opened 2017-10-23
feat(@schematics/angular): service worker project generation

alxhub:
This adds a feature to the Angular application schematic to generate a
project which includes @angular/service-worker's ServiceWorkerModule,
conditionally enabled for production builds. An ngsw.json file with
configuration suited for most projects is also generated.

Both operations are gated on the --service-worker flag which is added.


webmaxru:
It would be cool to have the configuration file name different from control file name, for example `ngsw-config.json`

PR #236, opened 2017-10-24
Purify improvements

filipesilva:


trotyl:
I guess this would close #102 as well?

filipesilva:
@trotyl not completely, but partially it does. Once `Purify` is gone then #102 is fixed. But it's a step forward.

trotyl:
That's great, thanks for the explanation and effort. 😃

PR #239, opened 2017-10-26
feat(typescript): allow downlevelIteration in Angular 5

cyrilletuzi:
Before TypeScript 2.3, when targeting ES5 (which is the case in Angular CLI), `for ... of` was limited to `Array`. Meaning it's not possible to iterate easily on new iterables like `Map`.

It's now possible since TypeScript 2.3, but requires the flag `downlevelIteration`. As Angular 5 requires TS >= 2.4, it's OK to enable this flag by default.

@hansl @filipesilva @Brocco I don't know how to manage the option to be Angular >= 5 (or TypeScript >= 2.3) only in schematics. Is there a variable available for that, or is there another system to separate between Angular versions (as I suppose similar cases will happen) ?

### DIFF
--- a/packages/schematics/angular/application/files/tsconfig.json
+++ b/packages/schematics/angular/application/files/tsconfig.json
@@ -14,6 +14,7 @@
     "lib": [
       "es2017",
       "dom"
-    ]
+    ],
+    "downlevelIteration": true
   }
 }


trotyl:
`downlevelIteration` would cause severe performance issue in generated file size, it's not suggested in TypeScript itself and has been [manually disabled](https://github.com/angular/zone.js/pull/916) in Zone.js.

cyrilletuzi:
> it's not suggested in TypeScript itself 

Can you provide a link about that ? [Official release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html) don't say it's not suggested.

About performance, the zone.js issue just says it produces a bigger bundle, not that it has performance impact. But : 
- the size issue it can be tempered by tslib `--importHelpers`, which I think Angular CLI is using ;
- it would be an issue just for ES5 bundles, ES2015 bundles (which is more and more becoming the new default) are not affected (but on contrary, they are currently affected by not being able to do what is possible in ES6 because of the current config).

And it's not the same for a complex library like zone.js to not add unnecessary tools for a few lines, and for a full app development to limit and complexify the dev experience for all a team.

trotyl:
From https://github.com/Microsoft/TypeScript/issues/3164#issuecomment-328587113. Not sure if it's an official point.

cyrilletuzi:
Closing as `forEach` is enough for `Map` and `Set` and DOM iteration irrelevant in Angular

PR #245, opened 2017-10-29
feat(@angular-devkit/core): add support for color and no-color

hansl:
And namespace the terminal.


packages/angular_devkit/core/src/terminal/caps.ts
```diff
@@ -0,0 +1,186 @@
...
+  return {
+    readable: stream.readable,
+    writable: stream.writable,
+    text: true,
+
+    colors: level > 0,
+    color256: level > 1,
+    color16m: level > 2,
+
+    rows: isTerminalStream ? _getRows() : null,
```
filipesilva: I think you can use `stream.isTTY` instead of passing in `isTerminalStream`: https://nodejs.org/api/tty.html.

We had a PR for doing that in the CLI recently https://github.com/angular/angular-cli/pull/8501

hansl: I've set the default to it, but technically this could be used outside of Node. I've let that other PR in, that's actually nice to have.

PR #254, opened 2017-11-02
feat(@schematics/angular): Add universal schematic

Brocco:


packages/schematics/angular/universal/files/__sourceDir__/__appDir__/__rootModuleFileName__
```diff
@@ -0,0 +1,14 @@
...
+import { AppComponent } from './app.component';
+
+@NgModule({
+  imports: [
+    AppModule,
+    ServerModule,
+  ],
+  bootstrap: [AppComponent],
+})
+export class AppServerModule {}
```
filipesilva: This should be `__rootModuleFileName__` instead of `AppServerModule`.

packages/schematics/angular/universal/index.ts
```diff
@@ -0,0 +1,215 @@
...
+      ...clientApp,
+      platform: 'server',
+      root: options.root,
+      outDir: options.outDir,
+      index: options.index,
+      main: options.main,
+      test: options.test,
+      tsconfig: options.tsconfigFileName,
+      testTsconfig: options.testTsconfigFileName,
+      environmentSource: 'environments/environment.ts',
```
filipesilva: `environments` should be used together with `environments`, or not at all.

Brocco:
I'm addressed the concerns expressed by @filipesilva

Toxicable:
Would it be possible to augment the main.ts to include 
```
document.addEventListener('DOMContentLoaded', () => {
  platformBrowserDynamic().bootstrapModule(AppModule);
});
```
This is essential for StateTransfer which is a likely addition to Universal and often overlooked/unknown

vikerman:
Opened https://github.com/angular/angular/issues/20484 to investigate how to avoid the race condition in TransferState without needing specific way to bootstrap the client side. We should make TransferState work more transparently to the user.

packages/schematics/angular/universal/index.ts
```diff
@@ -152,10 +105,45 @@ function findBrowserModuleImport(host: Tree, modulePath: string): ts.Node {
...
+      if (currentCall.kind === ts.SyntaxKind.ExpressionStatement) {
+        bootstrapCallExpression = currentCall;
+      }
+    }
+    bootstrapCall = currentCall;
+
+    const recorder = host.beginUpdate(mainPath);
+    const beforeText = `document.addEventListener('DOMContentLoaded', () => {\n  `;
+    const afterText = `\n});`;
+    recorder.insertRight(bootstrapCall.getStart(), beforeText);
```
hansl: Technically `insertLeft` ;)

PR #253, opened 2017-11-02
fix(@schematics/angular): update lint tools and rules for Angular 5

cyrilletuzi:
Codelyzer 4 is required for Angular 5 and some rules have been removed or deprecated in last versions of codelyzer and tslint.

Fixes angular/angular-cli#8324

@hansl @filipesilva

PR #257, opened 2017-11-03
feat: update codelyzer and add rules

mgechev:
codelyzer@^4.0.0 is out 🎉! It introduces support for `@angular/compiler` version 5 and few new rules such as:

- `angular-whitespace` - checks whether we have presence of whitespace
in the template to make it more readable, for instance
in `{{ expr }}`, `check-interpolation` will check if `expr` is
surrounded by whitespace. This rule also introduces auto fixes.
- `no-output-on-prefix` - warns if outputs are prefixed with `on` this
is based on
https://angular.io/guide/styleguide#dont-prefix-output-properties.

Tell me what you think about `angular-whitespace` and let me know if it
is too opinionated. As part of `angular-whitespace` we're planning a few
more features.

Also, since `invoke-injectable` is supposed to be caught by `tsc`, this rule has been deprecated.

Fixes angular/angular-cli#8463

### DIFF
--- a/packages/schematics/angular/application/files/package.json
+++ b/packages/schematics/angular/application/files/package.json
@@ -32,7 +32,7 @@
     "@types/jasmine": "~2.5.53",
     "@types/jasminewd2": "~2.0.2",
     "@types/node": "~6.0.60",
-    "codelyzer": "~3.2.0",
+    "codelyzer": "^4.0.1",
     "jasmine-core": "~2.6.2",
     "jasmine-spec-reporter": "~4.1.0",
     "karma": "~1.7.0",
--- a/packages/schematics/angular/application/files/tslint.json
+++ b/packages/schematics/angular/application/files/tslint.json
@@ -127,6 +127,8 @@
       "<%= prefix %>",
       "kebab-case"
     ],
+    "angular-whitespace": [true, "check-interpolation"],
+    "no-output-on-prefix": true,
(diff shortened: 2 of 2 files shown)


cyrilletuzi:
The update of codelyzer, along with a tslint one, is already proposed in #253

I don't think it's a good idea to enable the new `angular-whitespace` by default. Existing project would now fail on lint, and it's more a user preference.

akkumar:
The following commit message appears in the failed log. 

```console
    The following commit should always have a scope:
      1471daf feat: update codelyzer and add rules
  1 commits were found invalid...
```
May be - this needs an 'amend'ed commit again with the right syntax with the component name in it ?

say - "feat(aio): update codelyzer and add rules"

Would be good to have this fix in . 

cc @mgechev

moritz-h:
> I don't think it's a good idea to enable the new angular-whitespace by default. Existing project would now fail on lint, and it's more a user preference.

Yes, I found this discussion, right because of this behaviour. :smile: 

@mgechev 
When looking at https://angular.io/guide/styleguide there is not a single `{{ expr }}`. Everything is written as `{{expr}}`.

mgechev:
Yes, the style guide doesn't specify a recommended way of using the interpolation syntax. The fact that we're using `{{expr}}` doesn't mean that this is our recommendation. We don't want to introduce syntax conventions.

PR #258, opened 2017-11-04
feat(@schematics/angular): remove deprecated http package

cexbrayat:
Now that `@angular/http` is officially deprecated, we can probably remove it from the default package.json.

### DIFF
--- a/packages/schematics/angular/application/files/package.json
+++ b/packages/schematics/angular/application/files/package.json
@@ -17,7 +17,6 @@
     "@angular/compiler": "^5.0.0",
     "@angular/core": "^5.0.0",
     "@angular/forms": "^5.0.0",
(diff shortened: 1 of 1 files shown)


cyrilletuzi:
`@angular/platform-server`, which soon will be included by an universal schematic, still requires `@angular/http` to respect the depreciation cycle (see angular/angular#19788).

So it may be removed by default to avoid confusion, but then #254 should install it.

cexbrayat:
Closing as @Brocco made another PR and merged it in (#273)

Brocco:
My bad... I hadn't noticed this PR 🤭... also my PR was reverted as HTTP is still being used internally... it will get removed when those other places are cleaned up.

PR #260, opened 2017-11-05
refactor(@angular-devkit/build-optimizer): skip search for empty pure import list

clydin:


filipesilva:
Thanks for optimizing this 👍

PR #263, opened 2017-11-07
fix(@schematics/angular): ignore system files in service worker

cyrilletuzi:
Fixes angular/mobile-toolkit#180

When copying the assets, the CLI includes system files (e.g. .DS_Store and Thumbs.db). As a consequence, they are included in the service worker manifest as files to cache. But such file may (and should) not be deployed to the server, then a file will be missing, then the service worker installation will fail and so all the offline feature won't work.

This PR ensure the service worker will ignore these system files.

@hansl @filipesilva

### DIFF
--- a/packages/schematics/angular/application/files/__path__/ngsw-config.json
+++ b/packages/schematics/angular/application/files/__path__/ngsw-config.json
@@ -20,7 +20,9 @@
     "updateMode": "prefetch",
     "resources": {
       "files": [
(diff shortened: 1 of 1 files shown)


cyrilletuzi:
A better solution may be for the CLI to not copy those files when building.

cyrilletuzi:
Closing as replaced by a better solution in angular/angular-cli#8549

PR #264, opened 2017-11-07
fix(@schematics/angular): include fonts in service worker

cyrilletuzi:
Angular CLI does a great job including fonts from CSS in the build step (very useful in Material for example). The service worker should include them, otherwise the service worker will fail and the app won't work offline (when some ressources are missing, a service worker cancels installation).

I'm only including WOFF2 format, as all browsers supporting service workers also support WOFF2.

@hansl @filipesilva

### DIFF
--- a/packages/schematics/angular/application/files/__path__/ngsw-config.json
+++ b/packages/schematics/angular/application/files/__path__/ngsw-config.json
@@ -11,7 +11,8 @@
       "versionedFiles": [
         "/*.bundle.css",
         "/*.bundle.js",
-        "/*.chunk.js"
+        "/*.chunk.js",
+        "/*.woff2"
       ]
     }
   }, {


cyrilletuzi:
@alxhub

alxhub:
I don't think this change is needed. The default CLI app does not have `.woff2` files, so this line is meaningless unless the user adds them, at which point they should also change the `ngsw-config.json`. We cannot predict every type of file the user might want to cache or whether it's appropriate to prefetch them.

If the fonts are in `assets` they will be cached automatically by the `assets` group. That should be sufficient for the default case.

cyrilletuzi:
Makes sense.

Even if fonts are somewhere in between, as they are extracted by the CLI from the CSS, which themselves are cached by default (so if the user forgets to add fonts in cache config, the service worker installation will fail).

Then we should remember when a schematic will be added to add Material to modify the ngsw config accordingly, and be clear about that in the doc for other UI components libraries.

PR #267, opened 2017-11-08
fix(@schematics/angular): Correct imports of ViewEncapsulation

Brocco:
Fixes: angular/angular-cli#8374


JohannesHoppe:
Please don't forget this:
https://github.com/angular/devkit/blob/a8a6559649c8815b2b24561bd6ef00e4e4ad4ab8/packages/schematics/angular/application/files/__dot__angular-cli.json#L61

By default nothing is set for `viewEncapsulation` which yields to `None.`
I don't think that we want to have to have `None` by default. (should be `Emulated`)

see also:
https://github.com/angular/angular-cli/issues/8374

PR #268, opened 2017-11-10
fix(@schematics/angular): added alt tag to img

niklas-dahl:
get that 100/100 in Lighthouse Accessibility


niklas-dahl:
2017-11-10_10h20_43

current status

PR #270, opened 2017-11-10
fix(@schematics/angular): Update default encapsulation value.

Brocco:
Fixes: angular/angular-cli#8398


JohannesHoppe:
:tada:

rgantla:
I am currently running Cli @1.5.0 and updated my angular-cli.json file like below so as to achieve previous behavior where viewEncapsulation was emulated by default and no imports or properties were added to the component:
`"component": {
      "viewEncapsulation": "Emulated"
 }`

since this issue appears to be fixed, could you please let us know what version would include these changes? and if the newest version that has these changes, is out already so that I can update my cli?

PR #275, opened 2017-11-16
feat(@schematics/angular): custom suffix for component generation

leon:
Allow specifying a suffix for a component (Route, Dialog, Container).

```
ng g c —suffix=Route
```

When writing route components I find it useful to name all my route components `dashboard.route.ts` to signal that this is a component that is used by the router.
This makes it easy to filter out other components.

The same goes for `@angular/material` dialogs.
I want to name all dialogs with the Dialog suffix to easier filter out and find the relevant one.


leon:
@Brocco What do you think about this?
I tried to get something like this merged before we had schematics and you told me to wait :)

I'm not 100% sure about what the best name for the parameter. If you have a better suggestion I'm all ears.

alexeagle:
I think this is really an Angular style guide question.

PR #279, opened 2017-11-17
fix(@angular-devkit/build-optimizer): allow optimization of IIFE bundles

mgechev:
Rollup produces a bundle where it wraps everything into an IIFE. Because
of the `checkNodeForDecorators` check the scrub transformer was
returning before entering the IIFE and respectively wasn't dropping
Angular-specific decorators.

The current check enters the transform if the top-level node is not an
`ExpressionStatement` or is an `ExpressionStatement` and IIFE.

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
@@ -61,7 +61,9 @@ export function getScrubFileTransformer(program: ts.Program): ts.TransformerFact
       ts.forEachChild(sf, checkNodeForDecorators);
 
       function checkNodeForDecorators(node: ts.Node): void {
-        if (node.kind !== ts.SyntaxKind.ExpressionStatement) {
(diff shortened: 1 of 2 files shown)


IgorMinar:
Could this be a feature behind a flag. The rollup vs webpack usecases operate on very different kinds of code. With webpack we operate on sources while in rollup we operate on bundles. I'd prefer to keep these as explicit modes to prevent accidentally messing stuff up.

mgechev:
In fact, this feature is not required. A simple workaround would be to run the build optimizer on non-bundled files where there's no top-level IIFE.

Let me know if you'd still prefer to have it so I can push changes and introduce a flag. Otherwise, we can close the PR.

filipesilva:
@mgechev do you get the same bundle size reduction using that approach?

vyakymenko:
@filipesilva , yes we get a reduction.

filipesilva:
Ok that's good then.

PR #284, opened 2017-11-18
refactor(@schematics/angular): Removing Whitespace from default files

johnpapa:
I noticed that it reduces the lines of code and cognitive overload when teaching.

Also matches snippets from the style guide


hansl:
Closing this as this is not something we want (it increases the amount of changes when actually developing the app). I suggest coming up with custom schematics for teaching as this would make more sense (special casing for the uncommon use case).

PR #287, opened 2017-11-22
feat(@schematics/angular): Add appShell schematic

Brocco:


packages/schematics/angular/app-shell/index.ts
```diff
@@ -0,0 +1,292 @@
...
+      name: 'app-shell',
+      module: options.rootModuleFileName,
+    };
+
+    return schematic('component', componentOptions)(host, context);
+  };
+}
+
+export default function (options: AppShellOptions): Rule {
+  return (host: Tree, context: SchematicContext) => {
```
hansl: You don't need a function here, you can just `return chain([...]);`

packages/schematics/angular/universal/files/__sourceDir__/__appDir__/__rootModuleFileName__
```diff
@@ -7,7 +7,7 @@ import { AppComponent } from './app.component';
 @NgModule({
   imports: [
     AppModule,
-    ServerModule,
+    ServerModule
```
hansl: Why?

Brocco: The adding of imports logic does not support dealing with trailing commas being present yet.

hansl: that's a huge issue. Is this a problem only with this new logic?

packages/schematics/angular/collection.json
```diff
@@ -63,6 +63,11 @@
       "factory": "./universal",
       "description": "Create an Angular universal app.",
       "schema": "./universal/schema.json"
+    },
+    "appShell": {
```
hansl: `aliases: [ "app-shell" ]` ?

packages/schematics/angular/app-shell/index.ts
```diff
@@ -0,0 +1,292 @@
...
+    const app = getAppFromConfig(config, options.universalApp);
+    if (app === null) {
+      throw new SchematicsException('Universal/server app not found.');
+    }
+    const modulePath = getServerModulePath(host, app);
+    if (modulePath === null) {
+      throw new SchematicsException('Universal/server app not found.');
+    }
+    let moduleSource = getSourceFile(host, modulePath);
+    if (!isImported(moduleSource, 'Routes', '@angular/router')) {
```
jeffbcross: If I'm reading correctly, wouldn't this cause the new shell route not to be added if `Routes` had already been imported from `@angular/router`?

packages/schematics/angular/app-shell/index.ts
```diff
@@ -0,0 +1,292 @@
...
+  return modulePath;
+}
+// end helper functions.
+
+function addUniversalApp(options: AppShellOptions): Rule {
+  return (host: Tree, context: SchematicContext) => {
+    // Copy options.
+    const universalOptions = {
+      ...options,
+      name: options.universalApp,
```
jeffbcross: Should check for `universalApp` option and throw or set default?

packages/schematics/angular/app-shell/index.ts
```diff
@@ -0,0 +1,292 @@
...
+  const source = ts.createSourceFile(path, content, ts.ScriptTarget.Latest, true);
+
+  return source;
+}
+
+function getServerModulePath(host: Tree, app: AppConfig): string | null {
+  const mainPath = `/${app.root}/${app.main}`;
+  const mainSource = getSourceFile(host, mainPath);
+  const allNodes = getSourceNodes(mainSource);
+  const expNode = allNodes.filter(node => node.kind === ts.SyntaxKind.ExportDeclaration)[0];
```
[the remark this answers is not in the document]
Brocco: Agreed... this is what gets generated via the universal app.

Verifying the correct export would be a lot of logic, which should be added. Just not sure if that's required for the initial release.

packages/schematics/angular/app-shell/index.ts
Brocco: `universalApp` is required per the app-shell schematic. (may have been added after your review)

packages/schematics/angular/app-shell/index.ts
```diff
@@ -0,0 +1,292 @@
...
+  };
+}
+
+function addAppShellConfig(options: AppShellOptions): Rule {
+  return (host: Tree) => {
+    const config = getConfig(host);
+    const app = getAppFromConfig(config, options.clientApp || '0');
+
+    if (!app) {
+      throw new SchematicsException(`Clieng app (${options.clientApp}) could not be found.`);
```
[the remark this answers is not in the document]
Brocco: Slack typo fix notation appreciated 😍.

Consolidating and improving the messages is a good idea.

packages/schematics/angular/app-shell/index.ts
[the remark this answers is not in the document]
hansl: This can be done in the next RC, but is blocking the 1.6 final.

packages/schematics/angular/app-shell/index.ts
```diff
@@ -0,0 +1,292 @@
...
+          return (name as ts.StringLiteral).text === propertyName;
+      }
+
+      return false;
+    })[0];
+
+  return property as ts.PropertyAssignment;
+}
+
+function addRouterOutlet(options: AppShellOptions): Rule {
```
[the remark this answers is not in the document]
Brocco: `routerOutlet` was wrong, it was fixed between your review and this comment to use the correct `router-outlet` selector.

packages/schematics/angular/app-shell/index.ts
Brocco: You are reading that correctly. That logic could be more granular to check for existing routes outside of the import check.

Issue created: #290

packages/schematics/angular/app-shell/index.ts
```diff
@@ -0,0 +1,292 @@
...
+
+function addShellComponent(options: AppShellOptions): Rule {
+  return (host: Tree, context: SchematicContext) => {
+
+    const componentOptions = {
+      name: 'app-shell',
+      module: options.rootModuleFileName,
+    };
+
+    return schematic('component', componentOptions)(host, context);
```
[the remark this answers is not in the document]
Brocco: Yes that is correct, the DX of app shell should be discussed further.

PR #300, opened 2017-11-27
build: support building with TS 2.6

clydin:


package.json
```diff
@@ -72,11 +72,11 @@
...
     "ts-node": "^2.0.0",
     "tslint": "^5.5.0",
-    "typescript": "~2.4.0",
+    "typescript": "~2.6.1",
     "v8-profiler": "^5.7.0",
     "webpack-sources": "^1.0.1"
   },
   "devDependencies": {
-    "@bazel/typescript": "0.0.10"
+    "@bazel/typescript": "0.3.1"
```
filipesilva: This dependency is causing the bazel problems. Reverting it should fix it.

PR #303, opened 2017-11-27
fix(@angular-devkit/schematics): check that rules/sources return proper value

hansl:
Add a check if a rule returns an Observable, which will fail if the Observable does not
contain a Tree.


hansl:
Also always return an Observable, even in the case of an error.

PR #304, opened 2017-11-29
refactor(@angular-devkit/build-optimizer): optimize purify plugin

clydin:
This uses `exec` instead of `replace` and inserts only the pure annotation instead of a full replace.


filipesilva:
LGTM, thank you.

PR #310, opened 2017-12-04
fix(@schematics/angular): Use existing universal app if it exists.

### LINKED ISSUE #295: appShell does not reuse existing universal app
```
- [ ] feature request
```

### Area
```
- [ ] devkit
- [x] schematics
```

### Versions
```
Angular CLI: 1.6.0-rc.0
Node: 8.9.1
OS: darwin x64
Angular: 5.0.3
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 1.6.0-rc.0
@angular/platform-server: error
@angular-devkit/build-optimizer: 0.0.33
@angular-devkit/core: 0.0.21
@angular-devkit/schematics: 0.0.37
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.0-rc.0
@schematics/angular: 0.1.7
typescript: 2.4.2
```

### Repro steps

Generate a new app, run the universal schematic, than the appShell schematic:

```
ng new hello
ng g universal server
ng g appShell shell --universal-app server
```

This will fail as some files already exists:

```
error! src/app/app.server.module.ts already exists.
error! src/main.server.ts already exists.
error! src/tsconfig.server.json already exists.
```

If you remove this file, and run again `ng g appShell shell --universal-app server`,
then the command succeed, but `.angular-cli.json` contains two `server` app,
`BrowserModule` ends up with two `withServerTransition()`,
the bootstrap call is wrapped twice, etc...

### Desired functionality
It should probably reuse the existing universal app if it exists

Brocco:
Fixes #295

### DIFF
--- a/packages/schematics/angular/app-shell/index.ts
+++ b/packages/schematics/angular/app-shell/index.ts
@@ -66,6 +66,16 @@ function getServerModulePath(host: Tree, app: AppConfig): string | null {
 
 function addUniversalApp(options: AppShellOptions): Rule {
   return (host: Tree, context: SchematicContext) => {
+    const config = getConfig(host);
+    const appConfig = getAppFromConfig(config, options.universalApp);
+
+    if (appConfig && appConfig.platform === 'server') {
+      return host;
+    } else if (appConfig && appConfig.platform !== 'server') {
+      throw new SchematicsException(
+        `Invalid platform for universal app (${options.universalApp}), value must be "server".`);
+    }
+
     // Copy options.
     const universalOptions = {
       ...options,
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/app-shell/index.ts
```diff
@@ -66,6 +66,13 @@ function getServerModulePath(host: Tree, app: AppConfig): string | null {
 
 function addUniversalApp(options: AppShellOptions): Rule {
   return (host: Tree, context: SchematicContext) => {
+    const config = getConfig(host);
+    const appConfig = getAppFromConfig(config, options.universalApp);
+
+    if (appConfig && appConfig.platform === 'server') {
```
hansl: If `platform !== 'server'` but it exists, that should throw an exception.

Brocco: Additional check is added to throw for the invalid option value.

PR #312, opened 2017-12-05
fix(@schematics/angular): Apply the source dir option correctly for n…

Brocco:
…ew projects.

Fixes angular/angular-cli#8417


surjit:
when I can expect it fix to rollout to release ?

PR #320, opened 2017-12-07
feat(@schematics/schematics): add declaration files by default and np…

hansl:
…mignore


packages/schematics/schematics/blank/project-files/__dot__npmignore
```diff
@@ -0,0 +1,3 @@
+# Ignores TypeScript files, but keeps definitions.
+**/*.ts
+!**/*.d.ts
```
clydin: `**/` isn't needed.  It should be recursive unless prefixed with a `/`.

hansl: Done.

PR #325, opened 2017-12-07
fix(@schematics/angular): enable prod mode for universal

cyrilletuzi:
Universal build should be in production mode for performance.

It also avoid an issue with Material which produces an error (`window is not defined`) in development mode.

@filipesilva @hansl @Brocco


hansl:
Can you squash those two commits?

cyrilletuzi:
@hansl done

cyrilletuzi:
@Brocco ?

filipesilva:
@Brocco can you review?

PR #326, opened 2017-12-07
fix(@schematics/angular): fix for errors when content is null

DetectiveQuack:
Removed content attribute when not needed


clydin:
Superseded by https://github.com/angular/devkit/pull/300

PR #328, opened 2017-12-08
Virtual fs

hansl:


packages/angular_devkit/core/node/host.ts
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * An implementation of the Virtual FS using Node as the backend, synchronously.
+ */
+export class NodeJsSyncHost implements virtualFs.Host<fs.Stats> {
+  get capabilities(): virtualFs.HostCapabilities {
+    return { synchronous: true };
+  }
+
+  write(path: Path, content: virtualFs.FileBuffer): Observable<void> {
```
filipesilva: Feel a bit weird to have the sync host always return observables. Shouldn't it return what the observable is wrapping?

packages/angular_devkit/core/node/host_spec.ts
```diff
@@ -0,0 +1,61 @@
...
+import { normalize, virtualFs } from '@angular-devkit/core';
+import { NodeJsAsyncHost } from '@angular-devkit/core/node';
+import * as fs from 'fs';
+import { Observable } from 'rxjs/Observable';
+import { Subscription } from 'rxjs/Subscription';
+
+const temp = require('temp');
+
+
+describe('NodeJsAsyncHost', () => {
```
filipesilva: We also need tests for the sync host.

packages/angular_devkit/core/node/host_spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    let obs: Observable<virtualFs.HostWatchEvent>;
+    let subscription: Subscription;
+    const content = virtualFs.stringToFileBuffer('hello world');
+    const content2 = virtualFs.stringToFileBuffer('hello world 2');
+    const allEvents: virtualFs.HostWatchEvent[] = [];
+
+    Promise.resolve()
+      .then(() => fs.mkdirSync(root + '/sub1'))
+      .then(() => fs.writeFileSync(root + '/sub1/file1', 'hello world'))
+      .then(() => new Promise(resolve => setTimeout(resolve, 1000)))
```
filipesilva: On Angular CLI I've seen watchers register writes several seconds after they happen. Would not be surprised to have this flake out randomly. Also am not sure what the solution is.

packages/angular_devkit/core/node/host_spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  beforeEach(() => {
+    root = temp.mkdirSync('core-node-spec-');
+    host = new virtualFs.ScopedHost(new NodeJsAsyncHost(), normalize(root));
+  });
+  afterEach(done => {
+    host.delete(normalize('/'))
+      .subscribe({ complete() { done(); } });
+  });
+
+  it('can watch', done => {
```
filipesilva: Not a blocker for this PR but we'll need tests for the other methods. I can do those later.

packages/angular_devkit/core/src/virtual-fs/host/buffer.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { UnsupportedPlatformException } from '../../exception/exception';
+import { FileBuffer } from './interface';
+
+export function stringToFileBuffer(str: string): FileBuffer {
+  // If we're in Node...
```
filipesilva: This function only works for node and is only being used in specs. It's still part of `virtual-fs` so it would not be surprising if someone in the future imports it into non-spec code, making it node only.

A comment here would help prevent that, or maybe renaming the file to make it clear it's a spec helper.

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
```diff
@@ -0,0 +1,200 @@
...
+          }
+          subject.next({ path, time, type });
+
+          if (!options.persistent && type == HostWatchEventType.Deleted) {
+            subject.complete();
+            this._watchers.delete(currentPath);
+          }
+        });
+      }
+    } while (parent != currentPath);
```
filipesilva: Is notifying the parents always necessary? I can imagine foregoing this functionality for performance reasons. 

Maybe it could be an option in `HostWatchOptions`? Not sure all hosts would support it.

packages/angular_devkit/core/src/virtual-fs/host/interface.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+import { Observable } from 'rxjs/Observable';
+import { Path, PathFragment } from '../path';
+
+
+export declare type FileBuffer = ArrayBuffer;
+
+export interface HostWatchOptions {
+  readonly persistent?: boolean;
+  readonly recursive?: boolean;
```
filipesilva: We'll probably need polling as well in the future (Docker etc).

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
```diff
@@ -0,0 +1,200 @@
...
+  isFile(path: Path): Observable<boolean> {
+    return ArrayObservable.of(this._cache.has(path));
+  }
+
+  stats(_path: Path): Observable<Stats<{}>> | null {
+    return null;
+  }
+
+  watch(path: Path, options?: HostWatchOptions): Observable<HostWatchEvent> | null {
+    const subject = new Subject<HostWatchEvent>();
```
filipesilva: This way each new watcher gets a new subject. Maybe this is a stupid question, but isn't there a way to have a single subject with new watchers subscribing to it and filtering events by according to options?

packages/angular_devkit/core/node/host.ts
filipesilva: Is it meant to be used with `SyncDelegateHost` maybe?

hansl: Yes, it would be meant to be used with `SyncDelegateHost` if needed, but follow the `Host` interface so that it can be used where you need it.

packages/angular_devkit/core/node/host_spec.ts
hansl: There's no solution, this is at best a hack. The truth is I get like 5 events here, even with the `Sync` fs operations, on both Mac and Linux.

We can only verify that there is some event registered. Maybe I should watch first, then discard everything, then do some operations and check those. Let me experiment.

packages/angular_devkit/core/src/virtual-fs/host/buffer.ts
hansl: It's not just a spec helper. It will be used for example to create a `FileBuffer` from a string in schematics (since Tree methods can take both). I'm going to add the non-node version of this.

packages/angular_devkit/core/src/virtual-fs/host/interface.ts
hansl: Definitely, but let's get this in so we can use it.

packages/angular_devkit/core/src/virtual-fs/host/buffer.ts
hansl: BTW, the StackBlitz folks started using Schematics in the browser, so we already have a case for non-Node implementations. I'm going to start making it harder for us to add Node APIs outside of a `/node` sub-package.

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
hansl: Yes, because some watchers watch all subdirectories. I've added a check if there are no watchers to skip this.

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
hansl: We would have to get a new subject by matching watchers. I went down that path and it was too complicated. We should re-evaluate if it's a problem later.

filipesilva:
We're going to need a Windows CI soon to ensure this stuff really keeps working cross platform.

packages/angular_devkit/core/src/virtual-fs/path.ts
```diff
@@ -241,3 +241,25 @@ export function normalize(path: string): Path {
...
   }
 }
+
+
+// Platform-specific paths.
+export type WindowsPath = string & {
+  __PRIVATE_DEVKIT_WINDOWS_PATH: void;
+};
+export type PosixPath = string & {
+  __PRIVATE_DEVKIT_WINDOWS_PATH: void;
```
filipesilva: This should be `__PRIVATE_DEVKIT_POSIX_PATH` I think,

packages/angular_devkit/core/src/virtual-fs/host/sync.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Implement a synchronous-only host interface (remove the Observable parts).
+ */
+export class SyncDelegateHost<T extends object> {
+  constructor(protected _delegate: Host<T>) {
+    if (!_delegate.capabilities.synchronous) {
+      throw new SynchronousDelegateExpectedException();
+    }
+  }
+
+  protected _doSyncCall<ResultT>(observable: Observable<ResultT>): ResultT {
```
clydin: Not that this is necessarily bad but it seems like there's a decent amount of work going on here to get the result (sync call -> result -> observable -> _unwrap_ -> sync result).  I suppose that's unavoidable without a complete async/sync bifurcation.

packages/angular_devkit/core/node/host.ts
```diff
@@ -0,0 +1,342 @@
...
+  on(type: 'add', cb: (path: string) => void): ChokidarWatcher;
+  on(type: 'unlink', cb: (path: string) => void): ChokidarWatcher;
+
+  close(): void;
+}
+
+const { FSWatcher }: { FSWatcher: ChokidarWatcher } = require('chokidar');
+
+
+export type FsFunction0<R> = (cb: (err?: Error, result?: R) => void) => void;
```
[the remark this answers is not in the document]
hansl: Right. I thought they needed to be exported but only if `_callFs` is protected.

packages/angular_devkit/core/node/host.ts
```diff
@@ -0,0 +1,342 @@
...
+ * synchronous and one asynchronous.
+ */
+export class NodeJsAsyncHost implements virtualFs.Host<fs.Stats> {
+  private _callFs<R>(fn: FsFunction0<R>): Observable<R>;
+  private _callFs<R, T1>(fn: FsFunction1<R, T1>, p1: T1): Observable<R>;
+  private _callFs<R, T1, T2>(fn: FsFunction2<R, T1, T2>, p1: T1, p2: T2): Observable<R>;
+  private _callFs<R, T1, T2, T3>(fn: FsFunction3<R, T1, T2, T3>,
+                                 p1: T1, p2: T2, p3: T3): Observable<R>;
+
+  private _callFs<ResultT>(fn: Function, ...args: {}[]): Observable<ResultT> {
```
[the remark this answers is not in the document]
hansl: Could be, but I'd rather not. I removed it from the class though.

packages/angular_devkit/core/node/host.ts
```diff
@@ -0,0 +1,342 @@
...
+      if (fs.existsSync(this._getSystemPath(path))) {
+        return;
+      }
+      if (dirname(path) === path) {
+        throw new Error();
+      }
+      _createDir(dirname(path));
+      fs.mkdirSync(this._getSystemPath(path));
+    };
+    _createDir(dirname(path));
```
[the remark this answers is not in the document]
hansl: Good idea. Done.

packages/angular_devkit/core/src/virtual-fs/host/interface.ts
```diff
@@ -0,0 +1,68 @@
...
+  readonly time: Date;
+  readonly type: HostWatchEventType;
+  readonly path: Path;
+}
+
+export interface HostCapabilities {
+  synchronous: boolean;
+}
+
+export interface Host<StatsT extends object> {
```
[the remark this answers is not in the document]
hansl: Done.

packages/angular_devkit/core/src/virtual-fs/path.ts
hansl: Fixed thanks.

hansl:
@filipesilva @clydin PTAL, fixed every comments.

PR #334, opened 2017-12-14
feat(@angular-devkit/schematics): change schema validation to AJV

filipesilva:
Followup from https://github.com/angular/devkit/pull/280


packages/angular_devkit/schematics/src/tree/interface.ts
```diff
@@ -105,6 +105,13 @@ export interface Tree {
 }
 
 
+namespace Tree {
+  export function isTree(maybeTree: object): maybeTree is Tree {
```
filipesilva: This doesn't seem to be used.

packages/angular_devkit/schematics/tools/ajv-option-transform_spec.ts
```diff
@@ -0,0 +1,73 @@
...
+      .mergeMap(validator => validator(data))
+      .map(result => {
+        expect(result.success).toBe(true);
+        expect(data.obj.num).toBeUndefined();
+        expect(data.tslint).not.toBeUndefined();
+      })
+      .subscribe(done, done.fail);
+    });
+
+  it('works synchronously', done => {
```
filipesilva: There should be a way to fail synchronously if it's not sync. Otherwise you can't really know if it is sync or async and need to always prepare for it to be async.

packages/angular_devkit/schematics/tools/index.ts
```diff
@@ -13,3 +13,6 @@ export { FallbackEngineHost } from './fallback-engine-host';
 export { FileSystemEngineHost } from './file-system-engine-host';
 export { NodeModulesEngineHost } from './node-module-engine-host';
 export { NodeModulesTestEngineHost } from './node-modules-test-engine-host';
+
+export { AjvSchemaRegistry } from './ajv-option-transform';
```
filipesilva: I think we're going to have to move this to core if we want to share code with Build Facade or other things.

scripts/test.ts
```diff
@@ -25,6 +25,11 @@ require('source-map-support').install({
 });
 
 
+interface CoverageLocation {
```
filipesilva: This should be part of the `test: add an istanbul bootstrap script` commit.

packages/angular_devkit/schematics/src/tree/interface.ts
hansl: I think it should be where useful (instead of using TreeSymbol in Tree). It also communicates the types better to TypeScript. It doens't have to be part of this PR.

packages/angular_devkit/schematics/tools/ajv-option-transform_spec.ts
hansl: You can add the check, but this is used internally by ourselves so I don't see the need.

packages/angular_devkit/schematics/tools/index.ts
hansl: Yes.

scripts/test.ts
hansl: It should. I was just pushing stuff as quickly as I could :( sorry. I normally clean up before submitting for review.

packages/angular_devkit/schematics/tools/index.ts
hansl: Please do this tomorrow :)

scripts/test.ts
filipesilva: I was leaving the comment here mostly for myself when reworking this PR. I'll move it to the other commit.

packages/angular_devkit/schematics/tools/ajv-option-transform_spec.ts
filipesilva: Added a comment with this information.

packages/angular_devkit/schematics/src/formats/path.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import { normalize, schema } from '@angular-devkit/core';
+
+
+export const pathFormat: schema.SchemaFormat = {
+  name: 'path',
+  formatter: {
+    async: false,
+    validate: (path: string) => {
```
hansl: We should normalize the path instead of verifying if it's normalized already.

filipesilva: We can't do that in a format though. It just verifies that the data is following a format, it can't modify it.

packages/angular_devkit/schematics/src/formats/app-name.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { schema } from '@angular-devkit/core';
+import { htmlSelectorRe } from './html-selector';
+
+
+const unsupportedProjectNames = ['test', 'ember', 'ember-cli', 'vendor', 'app'];
```
clydin: Should the ember references be removed?  The existing CLI logic has them but that was really more of a carryover from the original ember code.

filipesilva: Done

hansl:
CLA: All authors of this PR have signed the CLA, so I'm overriding it.

PR #336, opened 2017-12-15
@schematics/update

hansl:


packages/angular_devkit/schematics/testing/schematic-test-runner.ts
```diff
@@ -9,6 +9,7 @@ import { logging, schema } from '@angular-devkit/core';
 import {
   Collection,
   DelegateTree,
+  Rule, Schematic, SchematicContext,
```
filipesilva: Shouldn't these be one per line?

packages/schematics/package-update/_angular-devkit/index.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Rule } from '@angular-devkit/schematics';
+import { SchematicsUpdateSchema } from '../schema';
+import { updatePackageJson } from '../utility/npm';
+
+
+const angularPackagesName = [
```
filipesilva: Can these always be updated to `@latest` and be compatible together? I don't think they can all be updated to a specific version (e.g. `0.1.2`) at least.

packages/schematics/package-update/_angular/index.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Rule } from '@angular-devkit/schematics';
+import { SchematicsUpdateSchema } from '../schema';
+import { updatePackageJson } from '../utility/npm';
+
+
+const angularPackagesName = [
```
filipesilva: This list shouldn't have any package names with 3+ parts, like `@*/*/testing` or `@angular/platform-browser/animations`. Those aren't real package names.

packages/schematics/package-update/_angular_cli/index.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Rule } from '@angular-devkit/schematics';
+import { SchematicsUpdateSchema } from '../schema';
+import { updatePackageJson } from '../utility/npm';
+
+
+export default function(options: SchematicsUpdateSchema): Rule {
+  return updatePackageJson(['@angular/cli'], options.version, options.loose);
```
filipesilva: Should updating Angular CLI itself should also update installed related tooling? Like Protractor and Karma versions.

packages/schematics/package-update/utility/npm.ts
```diff
@@ -0,0 +1,266 @@
...
+  'optionalDependencies',
+];
+
+
+const npmPackageJsonCache = new Map<string, Observable<JsonObject>>();
+
+function _getVersionFromNpmPackage(json: JsonObject, version: string, loose: boolean): string {
+  const distTags = json['dist-tags'] as JsonObject;
+  if (distTags && distTags[version]) {
+    return (loose ? '^' : '') + distTags[version] as string;
```
filipesilva: Why is `^` used for `loose` here but `~` below?

packages/schematics/package-update/utility/npm.ts
```diff
@@ -0,0 +1,266 @@
...
+  'dependencies',
+  'devDependencies',
+  'peerDependencies',
+  'optionalDependencies',
+];
+
+
+const npmPackageJsonCache = new Map<string, Observable<JsonObject>>();
+
+function _getVersionFromNpmPackage(json: JsonObject, version: string, loose: boolean): string {
```
filipesilva: I don't think this supports more complicated npm versions like `> 1.0.0 <= 1.2.3`. A comment saying this support is missing would be useful.

packages/schematics/package-update/utility/npm.ts
```diff
@@ -0,0 +1,266 @@
...
+ * @param {string} packageName The package name to fetch.
+ * @param {LoggerApi} logger A logger instance to log debug information.
+ * @returns {Observable<JsonObject>} An observable that will put the pacakge.json content.
+ * @private
+ */
+function _getNpmPackageJson(
+  packageName: string,
+  logger: logging.LoggerApi,
+): Observable<JsonObject> {
+  const url = `http://registry.npmjs.org/${packageName.replace(/\//g, '%2F')}`;
```
filipesilva: This won't work in environments with that need proxies for npm (like in China iirc). I don't know what the solution is for that but would like to bring it up.

packages/schematics/package-update/utility/npm.ts
```diff
@@ -0,0 +1,266 @@
...
+      if (!packages[depName] || packages[depName] === depVersion) {
+        return Observable.empty();
+      }
+      if (allVersions[depName] && semver.intersects(allVersions[depName], depVersion)) {
+        allVersions[depName] = semverIntersect.intersect(allVersions[depName], depVersion);
+
+        return Observable.empty();
+      }
+
+      return _getNpmPackageJson(depName, logger).pipe(
```
filipesilva: Should this be cached somewhere to prevent recursive calls from making a lot of repeat http calls? I'm not sure how deep this recursion goes.

packages/angular_devkit/schematics/src/engine/collection.ts
```diff
@@ -21,4 +21,8 @@ export class CollectionImpl<CollectionT extends object, SchematicT extends objec
   createSchematic(name: string): Schematic<CollectionT, SchematicT> {
     return this._engine.createSchematic(name, this);
   }
+
+  listSchematics(): string[] {
```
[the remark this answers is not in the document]
hansl: Thanks, I'll rename to `listSchematicNames` as it's clearer.

packages/angular_devkit/schematics/testing/schematic-test-runner.ts
hansl: Fixed. I blame webstorm.

packages/schematics/package-update/_angular-devkit/index.ts
hansl: They should all be to latest. I don't think we should talk about Angular 2, so I added a check that the version is at least 4.

filipesilva: These are actually the devkit (the variable should probably be `devkitPackagesName`). So even though they can be latest, they won't be able to be all at say `0.0.40` (that version exists for `@angular-devkit/schematics` but not `@angular-devkit/build-optimizer` for instance. 

`@angular-devkit/core` is an odd one to be in this list as well, since it's a dependency of the others and you'd probably not want to list it manually.

packages/angular_devkit/schematics/bin/schematics.ts
```diff
@@ -154,7 +154,7 @@ if (collection === null) {
 
 /** If the user wants to list schematics, we simply show all the schematic names. */
 if (argv['list-schematics']) {
-  logger.info(engineHost.listSchematics(collection).join('\n'));
+  logger.info(engine.listSchematicNames(collection).join('\n'));
```
Brocco: This should be part of the 2nd commit in the PR.

packages/schematics/package-update/_angular-devkit/index.ts
hansl: Yes, that's right. I've renamed the variable.

Those are automatic in the sense that you can run `ng update @angular-devkit` and it should work like you expect it (using latest). It's true that this wouldn't work for the devkit, since versions are not fixed. I'd like to fix that in the future, probably, maybe?

I've changed the devkit one to only use a tag.

packages/schematics/package-update/_angular/index.ts
hansl: Removed.

packages/schematics/package-update/_angular_cli/index.ts
hansl: Those should be separate packages.

packages/schematics/package-update/utility/npm.ts
hansl: `maxSatisfying` does support those.

packages/schematics/package-update/utility/npm.ts
hansl: Changed back to that.

packages/schematics/package-update/utility/npm.ts
hansl: No solution for that. Use some other tooling for now.

packages/schematics/package-update/utility/npm.ts
hansl: It is. See `npmPackageJsonCache`.

packages/angular_devkit/schematics/bin/schematics.ts
hansl: Oh right. Updated.

PR #339, opened 2017-12-17
refactor(@schematics/angular): Import dasherize from core

croso-df:
Seems like this `dasherize` is the same function exported on `@angular-devkit/core/src/utils/strings.ts`: https://github.com/angular/devkit/blob/master/packages/angular_devkit/core/src/utils/strings.ts#L46


croso-df:
Closing this one. Will open from my personal account.

PR #340, opened 2017-12-17
refactor(@schematics/angular): Import dasherize from core

caroso1222:
Seems like this dasherize is the same function exported on @angular-devkit/core/src/utils/strings.ts: https://github.com/angular/devkit/blob/master/packages/angular_devkit/core/src/utils/strings.ts#L46


hansl:
It is, but I'd like to scope those functions. Could you do that as part of this PR?

Basically, in core: 
```
import * as strings from './strings';

export { strings };
```

then:

```
import { strings } from '@angular-devkit/core';

...
strings.dasherize(...)
```

What do you think?

caroso1222:
@hansl I like the idea of scoping the string utils. I see that this stringUtils are being used all the way in pretty much all the angular/schematics. I'm happy to refactor all them to use the new scoped utility, but I better ask first. You want me to go through them real quick?

hansl:
@caroso1222 Yes please :)

caroso1222:
I rebased my branch with the latest master as it seemed to had changed quite significantly and I messed up the PR. Closing this and opening another one. Sorry about that.

PR #344, opened 2017-12-21
Schematic Tasks

clydin:


packages/angular_devkit/schematics/src/engine/task.ts
```diff
@@ -0,0 +1,117 @@
...
+    return Observable.from(this._queue.toArray());
+  }
+
+}
+
+// Naive priority queue
+class PriorityQueue<T> {
+  private _items = new Array<T>();
+
+  constructor(private comparator: (x: T, y: T) => number) {}
```
hansl: private members start with `_`.

packages/angular_devkit/schematics/src/engine/task.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  finalize(): Observable<TaskId> {
+    return Observable.from(this._queue.toArray());
+  }
+
+}
+
+// Naive priority queue
+class PriorityQueue<T> {
```
hansl: Could you move this to `core`? Seems like something we could use somewhere else.

packages/angular_devkit/schematics/src/engine/engine.ts
```diff
@@ -140,4 +162,42 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
...
+    }
+
+    return Promise.reject(new Error(`Unknown task '${name}' requested.`));
+  }
+
+  executeTasks(): Observable<void> {
+    const executors = new Map<string, TaskExecutor>();
+
+    const taskObservable = Observable
+      .from(this._taskSchedulers)
```
hansl: We use lettable operators now because it's where RxJS is going.

packages/angular_devkit/schematics/src/engine/engine.ts
```diff
@@ -140,4 +162,42 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
         return hostSource;
     }
   }
+
+  createTaskExecutor(name: string): Promise<TaskExecutor> {
+    // Hard coded tasks with default factory options for now
+    switch (name) {
+      case 'npm':
```
hansl: I think this line should be from a list of available task executors provided by the engine host.

packages/angular_devkit/schematics/tasks/npm_task.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {
+  TaskConfiguration,
+  TaskConfigurationGenerator,
+  TaskExecutor,
+} from '@angular-devkit/schematics';
+import { spawn } from 'child_process';
```
hansl: We should probably be able to split the task that schematics use with the implementation the host uses. Can you do that in this PR?

packages/angular_devkit/schematics/bin/schematics.ts
```diff
@@ -261,6 +261,7 @@ schematic.call(args, host, { debug, logger: logger.asApi() })
 
     return fsSink.commit(tree).ignoreElements().concat(Observable.of(tree));
   })
+  .concatMap(() => engine.executeTasks())
```
hansl: Use `executePostTasks`.

clydin: Done.

packages/angular_devkit/schematics/src/engine/engine.ts
clydin: Done.

packages/angular_devkit/schematics/src/engine/task.ts
clydin: Done.

packages/angular_devkit/schematics/src/engine/task.ts
clydin: Done.

packages/angular_devkit/schematics/tasks/npm_task.ts
clydin: Done.

packages/angular_devkit/schematics/tools/fallback-engine-host.ts
```diff
@@ -101,4 +103,27 @@ export class FallbackEngineHost implements EngineHost<{}, {}> {
 
     return [...allNames];
   }
+
+  async createTaskExecutor(name: string): Promise<TaskExecutor> {
```
hansl: We should use Observable everywhere.

packages/angular_devkit/schematics/src/engine/task.ts
```diff
@@ -0,0 +1,83 @@
...
+function compareTasks(x: TaskId, y: TaskId) {
+  if (x.priority === y.priority) {
+    return 0;
+  }
+
+  return x.priority > y.priority ? 1 : -1;
+}
+
+export class TaskScheduler {
+  private _queue = new PriorityQueue<TaskId>(compareTasks);
```
hansl: This seems like it could be simplified to `(x, y) => x.priority - y.priority`?

packages/angular_devkit/schematics/src/engine/task.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export type TaskExecutor<T = {}>
+  = (options: T | undefined, context: SchematicContext) => Observable<void>;
+
+export interface TaskExecutorFactory<T> {
+  readonly name: string;
+  create(options?: T): Promise<TaskExecutor>;
+}
+
+export interface TaskId {
```
hansl: Seems like the engine should keep a `Map<TaskId, TaskInfo>` cache and `TaskId` should just be the number. Otherwise everything needs to be readonly, which in this case it isn't.

packages/angular_devkit/schematics/tasks/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+import { TaskExecutorFactory } from '@angular-devkit/schematics';
+import { NodePackageTaskFactoryOptions } from './node-package/options';
+
+export { NodePackageInstallTask } from './node-package/install-task';
+export { NodePackageLinkTask } from './node-package/link-task';
+
+export class BuiltinTask {
+  static readonly NodePackage: TaskExecutorFactory<NodePackageTaskFactoryOptions> = {
+    name: 'node-package',
```
hansl: Could you export a list of names as well, so people can refer to the name as a token? Basically, if we rename `node-package` to something else we need tools to rename it as well, but if we export it they can just use the exported constant. Could be namespaced as well.

packages/angular_devkit/schematics/src/engine/task.ts
clydin: I kind of liked that the return type of the compare function was `1 | 0 | -1` but that's an arbitrary reason to keep it.

Done.

packages/angular_devkit/schematics/src/engine/task.ts
clydin: That was my original plan but was concerned about too many types.  But it does provide much better encapsulation.

Done.

packages/angular_devkit/schematics/tools/fallback-engine-host.ts
clydin: Done.

packages/angular_devkit/schematics/tasks/index.ts
clydin: `BuiltinTaskExecutor.NodePackage.name` will provide the name.  I also centralized the name inside `./node-packge/options.ts` so the string itself is only referenced once in the code.

PR #345, opened 2017-12-22
refactor: Update rxjs operators to be lettable.

Brocco:


Brocco:
Dependent upon angular/angular-cli#8976

packages/angular_devkit/core/node/cli-logger.ts
```diff
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import { logging, terminal } from '@angular-devkit/core';
-import 'rxjs/add/operator/filter';
+import { filter } from 'rxjs/operators';
```
CaerusKaru: This should be an import directly from `'rxjs/operators/filter'`

packages/angular_devkit/core/src/logger/transform-logger_spec.ts
```diff
@@ -6,21 +6,19 @@
  * found in the LICENSE file at https://angular.io/license
  */
 // tslint:disable:no-any
-import 'rxjs/add/operator/filter';
-import 'rxjs/add/operator/map';
+import { filter, map, toArray } from 'rxjs/operators';
```
CaerusKaru: These should be separate imports:

import { filter } from 'rxjs/operators/filter';
...

CaerusKaru: Per discussion in https://github.com/angular/angular-cli/pull/8976#discussion_r158592854, this is a non-starter/not necessary

PR #346, opened 2017-12-22
Build Optimizer Lightweight Transformations

clydin:
This provides a transformation fast path when source maps are not enabled and transformers that require the type checker are not needed for a file (`scrub-file`/`fold-file`).  This also has the advantage of not running any of the built-in (and unneeded) transformers when used.
There doesn't appear to be a publicly exposed mechanism to generate source maps with the TS printer.  If this were possible, the source map limitation could be lifted.

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer_spec.ts
+++ b/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer_spec.ts
@@ -88,17 +88,6 @@ describe('build-optimizer', () => {
       });
     });
 
-    it('doesn\'t process files without decorators/ctorParameters/outside Angular', () => {
-      const input = tags.oneLine`
-        var Clazz = (function () { function Clazz() { } return Clazz; }());
-        ${staticProperty}
-      `;
-
-      const boOutput = buildOptimizer({ content: input });
-      expect(boOutput.content).toBeFalsy();
-      expect(boOutput.emitSkipped).toEqual(true);
-    });
-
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer_spec.ts
```diff
@@ -88,7 +88,7 @@ describe('build-optimizer', () => {
       });
     });
 
-    it('doesn\'t process files without decorators/ctorParameters/outside Angular', () => {
+    xit('doesn\'t process files without decorators/ctorParameters/outside Angular', () => {
```
filipesilva: Why is this test removed?

clydin: The test should have been failing as `prefix-classes` will add the annotation but `testPrefixClasses` doesn't accept `var Clazz = (function () { function Clazz() { } return Clazz; }());`.

filipesilva: I see, this test isn't true anymore at all since we do process all files in some capacity... Maybe it's better to wholly remove the test instead of just doing `xit` then. What do you think?

It's probably true that processing all files is what's leading to slow builds though.

clydin: Removed the test as it is outdated now.

PR #354, opened 2018-01-03
fix(@schematics/angular): fix import path for generated flat modules

dkuerner:
If the target modulePath and the importModulePath are equal,
do not add a path seperator to the resulting module import path.

### DIFF
--- a/packages/schematics/angular/module/index_spec.ts
+++ b/packages/schematics/angular/module/index_spec.ts
@@ -55,6 +55,22 @@ describe('Module Schematic', () => {
     expect(files.indexOf('/projects/bar/src/app/foo/foo.module.ts')).toBeGreaterThanOrEqual(0);
   });
 
+  it('should import a generated flat module with a correct relative path', () => {
+    const options = {
+      ...defaultOptions,
+      appRoot: 'app',
+      module: 'app',
+      flat: true,
+    };
+
+    const tree = schematicRunner.runSchematic('module', options, appTree);
+    const files = tree.files;
+    expect(files.indexOf('/projects/bar/src/app/app.module.ts')).toBeGreaterThanOrEqual(0);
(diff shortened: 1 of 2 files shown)


dkuerner:
Fix for #355

packages/schematics/angular/module/index_spec.ts
```diff
@@ -42,6 +42,22 @@ describe('Module Schematic', () => {
...
+      module: 'app',
+      flat: true,
+    };
+
+    const tree = schematicRunner.runSchematic('module', options, appTree);
+    const files = tree.files;
+    expect(files.indexOf('/src/app/app.module.ts')).toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/src/app/foo.module.ts')).toBeGreaterThanOrEqual(0);
+    const moduleContent = getFileContent(tree, '/src/app/app.module.ts');
+    expect(moduleContent).toMatch(/import { FooModule } from '.\/foo.module'/);
```
Brocco: Please update regex to this:
```
/import { FooModule } from '\.\/foo.module'/
```
This way it checks for a period and not any character.

Brocco (changes requested):
One minor change.

packages/schematics/angular/module/index_spec.ts
dkuerner: Absolutely.. added and squashed the requested change

Brocco (dismissed):

filipesilva:
@dkuerner can you rebase please? I'm not sure how the CI error relates to your changes and suspect it will go away on a rebase.

alexeagle:
Sorry @dkuerner  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #358, opened 2018-01-04
build: update bazel WORKSPACE

clydin:


WORKSPACE
```diff
@@ -1,18 +1,23 @@
+workspace(name = "angular_devkit_src")
```
alexeagle: the `_src` suffix in angular/angular was because I had two workspaces in that repo, but I'm trying to move away from that. Just `angular_devkit` is better.
Note that the workspace name will appear in absolute imports, just as if you were deep-importing from under an npm package. That means that code is more portable: a file could be copied from inside the project to a dependent project and imports should stay the same.

WORKSPACE
```diff
@@ -1,18 +1,23 @@
...
 git_repository(
     name = "build_bazel_rules_nodejs",
     remote = "https://github.com/bazelbuild/rules_nodejs.git",
     tag = "0.3.1",
 )
 
-load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")
+load("@build_bazel_rules_nodejs//:defs.bzl", "check_bazel_version", "node_repositories")
+
+check_bazel_version("0.8.1")
```
alexeagle: bazel 0.9 is out now, and resolves the issue that caused us some trouble with rxjs getting included in `bazel build ...` so maybe we should require 0.9.0?

WORKSPACE
```diff
@@ -1,18 +1,23 @@
...
 node_repositories(package_json = ["//:package.json"])
 
 git_repository(
     name = "build_bazel_rules_typescript",
     remote = "https://github.com/bazelbuild/rules_typescript.git",
-    tag = "0.6.0",
+#    tag = "0.7.1+",
+    commit = "89d2c75066bea3d9c942f29dd1d2ea543c58d6d5"
```
alexeagle: this was needed for Angular to upgrade to TS 2.6. If possible it would be better to use a tag instead.

WORKSPACE
clydin: ngcontainer appears to only have 0.8.1

alexeagle:
Well I could easily push another ngcontainer with bazel 0.9 but then you'll
run into the issue Filipe reported. 0.8.1 will work for now.

clydin:
That issue appears to only impact the CLI proper.  I switched this PR over to `ngcontainer` 0.0.8 without issue.

PR #361, opened 2018-01-05
build: update to bazel 0.9

clydin:
use the new default tsconfig option as well.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -8,7 +8,7 @@ git_repository(
 
 load("@build_bazel_rules_nodejs//:defs.bzl", "check_bazel_version", "node_repositories")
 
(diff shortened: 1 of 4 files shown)


WORKSPACE
```diff
@@ -19,4 +19,4 @@ git_repository(
 
 load("@build_bazel_rules_typescript//:setup.bzl", "ts_setup_workspace")
 
-ts_setup_workspace()
+ts_setup_workspace(default_tsconfig = "@angular_devkit//:tsconfig.json")
```
alexeagle: FYI, this may have to change a bit, see https://github.com/bazelbuild/rules_typescript/issues/122

PR #362, opened 2018-01-06
fix(@schematics/angular): include support for ES7 proposals

markgoho-EDT:
Add support for ES7 proposals such as Object.values and Object.entries, fixes angular/angular-cli#9111


markgoho-EDT:
Dangit, submitted with the wrong github account.

PR #363, opened 2018-01-06
fix(@schematics/angular): Add support for ES2017 proposals

markgoho:
Add support for ES7 proposals such as Object.values and Object.entries, fixes angular/angular-cli#9111

### DIFF
--- a/packages/schematics/angular/application/files/__sourcedir__/polyfills.ts
+++ b/packages/schematics/angular/application/files/__sourcedir__/polyfills.ts
@@ -34,7 +34,7 @@
 // import 'core-js/es6/weak-map';
 // import 'core-js/es6/set';
 
-// Support for ES7 Proposals
+// Support for ES2017 Proposals
 // import 'core-js/stage/4';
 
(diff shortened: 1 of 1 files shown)


markgoho:
Sorry guys, I just can't seem to get the commit messages right. @filipesilva has tried to walk me through rebasing the commit but it never seems to work. Can someone handle that?

trotyl:
`Object#values` and `Object#entries` are ES8 (ES2017) features

markgoho-EDT:
@trotyl you're absolutely right, we should make a PR to Core-JS after this one is merged 👍 https://github.com/zloirock/core-js#ecmascript-7-proposals

markgoho:
@trotyl you're absolutely right, we should make a PR to Core-JS after this one is merged 👍 https://github.com/zloirock/core-js#ecmascript-7-proposals

clydin:
Thank you for the contribution.  However, as noted in the referenced issue, only polyfills required by Angular itself are intended to be included by default.

And in regards to the commit message, you almost had it right.  `schematics` and `angular` are backwards (should be `@schematics/angular`).

PR #366, opened 2018-01-10
feat(@angular-devkit/architect): add package

filipesilva:
Todos:
- [x] improve registry error message context (see https://github.com/angular/devkit/pull/366#discussion_r165707044)
- [x] remove duplication of base target options in build facade target schemas
- [x] remove build optimizer from root package.json
- [x] change `webpack-build-facade` to `build-webpack`
- [x] fix `build-webpack` dependencies
- [x] rename `build-facade-cli` to `architect`
- [x] use only `run` targets
- [x] make build system target a generic (see https://github.com/angular/devkit/pull/366#discussion_r165742535)
- [x] replace `& JsonObject` in `TargetOptions` with a generic with default
- [x] find a way to not repeat base target schema in build-webpack
- [x] figure out good way to cancel running targets
- [x] create new error instead of using ajv error array (see https://github.com/angular/devkit/pull/366#discussion_r165724800)
- [x] change `mode` to `optimizationLevel`
- [x] use `NodeJsSyncHost` 
- [x] clarify workspace as interface comment (see https://github.com/angular/devkit/pull/366#discussion_r165800137)

Blocked:
- [ ] setup postinstall step to install packages that will not be in google3
- [ ] clean up old CLI files
- [ ] change `environments` to `replaceFile`
- [ ] change `main/polyfills/etc` to `entryPoints`
- [ ] use `BuildEventProtocol` for build events (see https://github.com/angular/devkit/pull/366#discussion_r165724035)
- [ ] add e2e harness and remove peer dependencies from root package.json
- [ ] remove `ajv-async` and use node 8 in appveyor (see https://github.com/angular/devkit/pull/366#discussion_r165833706)
- [ ] further Angular CLI integration (https://github.com/angular/angular-cli/pull/9337)
- [ ] update schematics to use build facade workspace file
- [ ] add input and output VFS (also fixes https://github.com/angular/angular-cli/issues/7305)

Proposals:
- [x] replace `default: true` in project/target with `defaultProject/defaultTarget: string` in workspace/project
- [x] make `configurations` an array inside a target


packages/angular_devkit/build_facade/src/build-system.ts
```diff
@@ -0,0 +1,98 @@
...
+  flags: string[];
+}
+
+export interface BuildArtifact {
+  name: string;
+  outputs: BuildAsset[];
+  flags: string[];
+}
+
+export interface BuildEvent {
```
hansl: Look into the `BuildEventProtocol` from Bazel, and lets use that interface. Then the mapping will be easy and it works with any type system (and is extensible).

https://docs.bazel.build/versions/master/build-event-protocol.html

You can find the protos here: https://github.com/googleapis/googleapis/tree/master/google/devtools/build/v1

Can be done in a separate PR but this is required for serve since it relies on build.

packages/angular_devkit/build_facade/src/build-system.ts
```diff
@@ -0,0 +1,98 @@
...
+    const registry = new schema.CoreSchemaRegistry([],
+      this._getSystemPath(dirname(this.schemaPath)));
+
+    return registry.compile(workspaceSchema).pipe(
+      mergeMap(validator => validator(options)),
+      mergeMap(validatorResult => {
+        if (validatorResult.success) {
+          return of(options);
+        } else {
+          return _throw(validatorResult.errors);
```
hansl: `errors` is an array of strings. You should create a new error object since that's what's expected.

packages/angular_devkit/build_facade/src/build-system.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+      }),
+    );
+  }
+
+  private _getSystemPath(path: Path): string {
```
hansl: Using a `NodeJsSyncHost` would remove any path stuff from this whole PR.

packages/angular_devkit/build_facade/src/target.ts
```diff
@@ -0,0 +1,36 @@
...
+import { BaseException, JsonObject } from '@angular-devkit/core';
+
+
+export class MissingTargetBuildSystemException extends BaseException {
+  constructor(props: JsonObject) {
+    super(`Configuration does not contain or inherit build system: ${props}.`);
+  }
+}
+
+export class Target {
```
hansl: This should be an interface.

packages/angular_devkit/build_facade/src/architect_spec.ts
```diff
@@ -0,0 +1,248 @@
...
+      },
+    },
+  };
+
+  it('works', (done) => {
+    const workspace = new Workspace(normalize('/path/to/workspace'), props);
+    workspace.load().subscribe({
+      complete: () => {
+        const architect = new Architect(workspace);
+        const targetOptions = architect.createTargetOptionsFromWorkspace();
```
hansl: To me this feels like two different things;

1. workspace is just an interface, loaded from the `.angular.json`. There could be an `Architect.loadWorkspaceFromHost(host: Host, path: Path)` and `Architect.validateWorkspace(workspace: JsonObject): Workspace`.
2. Architect could be workspace agnostic. So the workspace isn't passed in the constructor, but to each methods.
3. Architect could take either a workspace and a list of names, or just a list of JSON objects to validate and build.

packages/angular_devkit/core/package.json
```diff
@@ -11,7 +11,9 @@
     "core"
   ],
   "dependencies": {
-    "ajv": "~5.5.1",
+    "ajv": "^6.1.0",
+    "ajv-async": "^1.0.1",
```
hansl: Question; if we limit to Node 8+ do we still need ajv-async?

filipesilva: We do not, no. But we'll need update appveyor to use node 8 as well.

package.json
```diff
@@ -45,38 +46,95 @@
   },
   "homepage": "https://github.com/angular/devkit",
   "dependencies": {
+    "@angular-devkit/build-optimizer": "0.0.41",
```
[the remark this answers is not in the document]
filipesilva: I'm pretty sure too. I left it in by accident while transferring over dependencies. Will remove.

package.json
```diff
@@ -45,38 +46,95 @@
...
+    "@angular/common": "^5.2.1",
```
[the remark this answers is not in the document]
filipesilva: We don't want them, no. But to testing the webpack build facade right now is done through unit tests instead of e2e tests, so we don't have a way to create peerDep setups. 

Once we have a way to setup e2e tests these can go away.

package.json
```diff
@@ -45,38 +46,95 @@
...
+    "ajv-merge-patch": "^3.0.0",
+    "autoprefixer": "^7.2.3",
     "chokidar": "^1.7.0",
+    "circular-dependency-plugin": "^4.3.0",
+    "clean-css": "^4.1.9",
+    "common-tags": "^1.5.1",
     "conventional-changelog": "^1.1.0",
+    "copy-webpack-plugin": "^4.2.3",
+    "css-loader": "^0.28.7",
+    "denodeify": "^1.2.1",
```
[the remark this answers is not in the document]
filipesilva: I agree. That should go away when cleaning up the CLI webpack files.

package.json
```diff
@@ -45,38 +46,95 @@
...
+    "exports-loader": "^0.6.4",
+    "extract-text-webpack-plugin": "^3.0.2",
+    "file-loader": "^1.1.5",
```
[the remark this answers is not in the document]
filipesilva: I think that's reasonable. Maybe just install them a the package level (in `packages/angular_devkit/webpack_build_facade/`) and let node resolution pick up the remaining packages from the top.

packages/angular_devkit/build_facade/src/architect.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    // TODO: should buildSystem and targetType be removed as well? Reconsider this design.
+    // Remove unneeded properties.
+    delete targetOptions['default'];
+    delete targetOptions['targets'];
+    delete targetOptions['configurations'];
+
+    // Resolve root for the target.
+    // TODO: add Path format to JSON schemas
+    targetOptions.root = resolve(this._workspace.path, normalize(targetOptions.root));
```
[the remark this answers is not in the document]
filipesilva: I think that makes sense for the workspace root but not for project paths. Those should be relative to the project root.

packages/angular_devkit/build_facade/src/build-system.ts
filipesilva: I agree this is how we should transmit build artifacts but at the moment `serve` needs a bit more than this. It actually needs the webpack compilation since it hooks into it and adds entry points for live reload.

e2e only needs to know when serve has emitted a build event, and doesn't care about the content.

I think what we want right now is to hook build and serve directly while we figure out the build events and dependency model. When we have that, maybe rework the dev server to be more generic.

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -119,7 +121,6 @@ export class CoreSchemaRegistry implements SchemaRegistry {
     }
 
     this._ajv = ajv({
-      removeAdditional: 'all',
```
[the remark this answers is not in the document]
filipesilva: Setting it to any other option but false also causes the validation to succeed instead of failing. Specifically, the `fails on invalid additionalProperties` test in `packages\angular_devkit\core\src\json\schema\registry_spec.ts` starts failing.

packages/angular_devkit/build_facade/src/target.ts
filipesilva: It should. I think you mean the `Target` class in `build_facade/src/build-system.ts` though.

Having these two is confusing. These simpler project/target/configuration inside the workspace should have different names or be removed altogether.

packages/angular_devkit/build_facade/src/architect_spec.ts
filipesilva: I don't quite get what you mean here. Can we talk about this?

.monorepo.json
```diff
@@ -45,6 +45,22 @@
       "version": "0.3.1",
       "hash": "d99e730f5b1c9e8b585fb44ee61987c6"
     },
+    "@angular-devkit/architect": {
```
hansl: Please name `architect-cli` and name the package `architect`.

packages/angular_devkit/architect/bin/architect.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  return null;
+}
+
+/**
+ * Show usage of the CLI tool, and exit the process.
+ */
+function usage(exitCode = 0): never {
+  logger.info(tags.stripIndent`
+    architect build|run [project][:target][:configuration] [options, ...]
```
hansl: We don't support `build` anymore, do we?

packages/angular_devkit/build_facade/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@angular-devkit/build-facade",
```
hansl: Move the files and rename the package to `architect`, like we do with `schematics`.

packages/angular_devkit/build_facade/src/architect.ts
```diff
@@ -0,0 +1,177 @@
...
+  targetType: string;
+}
+
+export interface TargetInfo<OptionsT = JsonObject> {
+  properties: TargetProperties;
+  options: OptionsT;
+}
+
+export class Architect {
+  _host: virtualFs.SyncDelegateHost<fs.Stats>;
```
hansl: Does it need to be `Sync`?  And do you need `fs.Stats`?  Using `fs.Stats` here limit this to Node only, and I don't see the stats method being used. You can use the `{}` typing if you don't care what it is.

packages/angular_devkit/build_facade/src/architect.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  constructor(private _workspace: Workspace) {
+    this._host = new virtualFs.SyncDelegateHost(new NodeJsSyncHost());
```
hansl: You should probably get the host from the caller. I know some targets won't have a host as input, but if they do we should always use the host provided.

packages/angular_devkit/build_facade/src/architect.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    return {
+      properties: targetProperties,
+      options: targetOptions as OptionsT,
+    };
+  }
+
+  // Will run the target (if the target supports run), using the targetInfo.
+  run<OptionsT>(
+    targetInfo: TargetInfo<OptionsT>, logger?: logging.Logger, stop?: Observable<void>,
```
hansl: Is the `stop` observable a normal pattern? I thought a better (and more used) pattern was to `cancel()` the returned observable instead.

packages/angular_devkit/build_facade/src/target.ts
```diff
@@ -0,0 +1,81 @@
...
+export interface BuildEvent {
+  success: boolean;
+}
+
+// TODO: need a way to show/return option help.
+export class Target<OptionsT> {
+  constructor(public schemaPath: Path, public context: TargetContext) { }
+
+  protected validateOptions(options: Partial<OptionsT>): Observable<OptionsT> {
+    const schemaContent = virtualFs.fileBufferToString(this.context.host.read(this.schemaPath));
```
hansl: We need to use host everywhere. SyncHost isn't compatible with a bunch of use cases.

packages/angular_devkit/build_facade/src/workspace-entities/configuration.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class MissingConfigurationBuildFacadeException extends BaseException {
+  constructor(props: JsonObject) {
+    super(`Configuration does not contain or inherit build facade: ${props}.`);
+  }
+}
+
+export class Configuration {
+  constructor(readonly name: string, readonly props: JsonObject) {
+    if (!this.props.buildFacade) {
```
hansl: Again, I think these (Workspace, Project, Target and Configuration) should be interfaces that are validated by the Architect instance against their respective Schema, but there's no reason to have `x.props.xyz` to access `xyz`.  Look at how I did it in Schematic for making it extensible (the definition types).

packages/angular_devkit/architect/bin/architect.ts
filipesilva: Correct, removing.

packages/angular_devkit/build_facade/src/architect.ts
filipesilva: Agreed.

packages/angular_devkit/build_facade/src/architect.ts
filipesilva: I got it from https://medium.com/@benlesh/rxjs-dont-unsubscribe-6753ed4fda87. Might be dated though.

I don't see any `cancel()` on RxJs though. Do you mean `.unsubscribe()`? I don't think that allows for the cleanup, and that article advises against `.unsubscribe()`.

packages/angular_devkit/build_facade/src/architect.ts
filipesilva: Done.

packages/angular_devkit/build_facade/src/target.ts
filipesilva: Done.

.monorepo.json
filipesilva: Fixed

packages/angular_devkit/build_facade/src/workspace-entities/configuration.ts
filipesilva: Yeah I agree their implementation isn't very good now. This was the first bit I made and it ended up not very good. Fixing.

hansl:
All your `run()` functions are over 200 lines. We need to split them up. Also note to self; add a lint rule for code complexity.

packages/angular_devkit/architect/src/builder.ts
```diff
@@ -0,0 +1,45 @@
...
+  success: boolean;
+}
+
+export interface Builder<OptionsT> {
+  // TODO: how do I actually specify the constructor signature?
+  // new(context: BuilderContext): Builder<OptionsT>;
+  run(_info: Target<Partial<OptionsT>>, _stop?: Observable<void>): Observable<BuildEvent>;
+}
+
+export interface BuildersDescription {
```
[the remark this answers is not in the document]
filipesilva: SGTM.

packages/angular_devkit/architect/src/workspace.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface WorkspaceProject {
+  projectType: 'application' | 'library';
+  root: string;
+  defaultTarget?: string;
+  targets: { [k: string]: WorkspaceTarget };
+}
+
+export interface WorkspaceTarget {
```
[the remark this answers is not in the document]
filipesilva: Done.

packages/angular_devkit/build_webpack/src/browser/index.ts
```diff
@@ -0,0 +1,259 @@
...
+import { requireProjectModule } from '../angular-cli-files/utilities/require-project-module';
+import {
+  statsErrorsToString,
+  statsToString,
+  statsWarningsToString,
+} from '../angular-cli-files/utilities/stats';
+const webpackMerge = require('webpack-merge');
+
+
+export interface BrowserBuilderOptions {
```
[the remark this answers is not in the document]
filipesilva: Done.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,262 @@
...
+        ...additionalInput ? additionalInput as {} : {},
+      } as OptionsT,
+    };
+
+    return target;
+  }
+
+  // Will run the target using the target.
+  run<OptionsT>(
+    target: Target<OptionsT>, logger?: logging.Logger, stop?: Observable<void>,
```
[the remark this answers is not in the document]
filipesilva: Done.

packages/angular_devkit/build_webpack/src/dev-server/index.ts
```diff
@@ -0,0 +1,376 @@
...
+        // https://github.com/angular/angular-cli/issues/7197
+        // https://github.com/nodejs/node/issues/13391
+        // https://github.com/nodejs/node/commit/2cb6f2b281eb96a7abe16d58af6ebc9ce23d2e96
+        if (/^v8.\d.\d+$/.test(process.version)) {
+          httpServer.keepAliveTimeout = 30000; // 30 seconds
+        }
+
+        if (stop) {
+          // Close the server and complete run if the stop observable completes.
+          stop.subscribe({
```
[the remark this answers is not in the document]
filipesilva: Thanks, this was the bit I was missing.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,262 @@
...
+        return of(this);
+      }),
+    );
+  }
+
+  getTarget<OptionsT>(
+    projectName?: string,
+    targetName?: string,
+    configurationName?: string,
+    additionalInput?: Partial<OptionsT>,
```
clydin: Should `additionalInput` be first or have some overloads?  With the current order it doesn't appear possible to augment the default project/target without looking up the names and supplying them.

As an alternative, could `additionalInput` be dropped?  The target options could be adjusted after the call.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,262 @@
...
+    };
+
+    return target;
+  }
+
+  // Will run the target using the target.
+  run<OptionsT>(
+    target: Target<OptionsT>, logger?: logging.Logger, stop?: Observable<void>,
+  ): Observable<BuildEvent> {
+    logger = logger || createConsoleLogger();
```
clydin: Should the default be a null logger instead of console?  Not specifying a logger, to me at least, would indicate I do not want logging.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,262 @@
...
+
+  getBuilderDescription<OptionsT>(target: Target<OptionsT>): Observable<BuilderDescription> {
+    return new Observable((obs) => {
+      // TODO: this probably needs to be more like NodeModulesEngineHost.
+      const basedir = getSystemPath(this._root);
+      const [pkg, builderName] = target.builder.split(':');
+      const pkgJsonPath = nodeResolve(pkg, { basedir, resolvePackageJson: true });
+
+      // Load builders json.
+      const pkgJsonBuildersPath = require(pkgJsonPath)['builders'];
```
clydin: nit. `builders` as a plural would generally imply an array or object map property.  This may cause some confusion.

filipesilva:
@hansl PTAL I think all the comments are addressed in the commits I just pushed. I only refactored the dev-server and protractor builders to be smaller though, all the other run functions were already within the ~90 lines.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,270 @@
...
+export interface Target<OptionsT = {}> {
+  root: Path;
+  projectType: string;
+  builder: string;
+  options: OptionsT;
+}
+
+export class Architect {
+  private _workspaceSchema = join(normalize(__dirname), 'workspace-schema.json');
+  private _buildersSchema = join(normalize(__dirname), 'builders-schema.json');
```
clydin: These two seem to never change; should either be readonly, elevated to module level constants, or class static properties (if we want developers to find/use the schemas; e.g., static readonly workspaceSchemaPath = ...).

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,270 @@
...
+      const [pkg, builderName] = target.builder.split(':');
+      const pkgJsonPath = nodeResolve(pkg, { basedir, resolvePackageJson: true });
+      let buildersJsonPath: Path;
+
+      // Read the `builders` entry of package.json.
+      return this._host.read(normalize(pkgJsonPath)).pipe(
+        concatMap(buffer =>
+          of(parseJson(virtualFs.fileBufferToString(buffer), JsonParseMode.Loose))),
+        concatMap((pkgJson: JsonObject) => {
+          const pkgJsonBuildersentry = pkgJson['builders'] as string;
```
clydin: nit. builders as a plural would generally imply an array or object map property. This may cause some confusion.

packages/angular_devkit/architect/src/builder.ts
```diff
@@ -0,0 +1,43 @@
...
+  builders: { [k: string]: BuilderDescription };
+}
+
+export interface BuilderDescription {
+  class: Path;
+  schema: Path;
+  description: string;
+}
+
+export interface BuilderClass<OptionsT> {
```
clydin: nit. `BuilderConstructor<OptionsT>`?

packages/angular_devkit/build_webpack/src/dev-server/index.ts
```diff
@@ -0,0 +1,376 @@
...
+  disableHostCheck?: boolean;
+}
+
+export class DevServerBuilder implements Builder<DevServerBuilderOptions> {
+
+  constructor(public context: BuilderContext) { }
+
+  run(target: Target<DevServerBuilderOptions>, stop?: Observable<void>): Observable<BuildEvent> {
+
+    const root = getSystemPath(target.root);
```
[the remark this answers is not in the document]
filipesilva: Done.

packages/angular_devkit/build_webpack/src/dev-server/index.ts
```diff
@@ -0,0 +1,376 @@
...
+    const options = target.options;
+
+    let browserOptions: BrowserBuilderOptions,
+      browserTarget: Target<BrowserBuilderOptions>;
+
+    return fromPromise(checkPort(options.port, options.host)).pipe(
```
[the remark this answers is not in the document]
filipesilva: Done.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,262 @@
...
+  builder: string;
+  options: OptionsT;
+}
+
+export class Architect {
+  private _workspaceSchema = pathResolve(__dirname, 'workspace-schema.json');
+  private _buildersSchema = pathResolve(__dirname, 'builders-schema.json');
+  private _workspace: Workspace;
+
+  constructor(private _root: Path, private _host: virtualFs.Host<{}>) { }
```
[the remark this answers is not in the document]
filipesilva: @hansl and I were talking about doing something slightly different to get rid of root: resolve all paths in the workspace config via json schema transforms.

packages/angular_devkit/architect/src/architect.ts
filipesilva: I was actually thinking there should be an options object to avoid these problem. WDYT?

packages/angular_devkit/architect/src/architect.ts
filipesilva: Hm... that's debatable IMHO. I think it would generally be surprising to not have any output with the defaults. @hansl WDYT?

packages/angular_devkit/architect/src/architect.ts
filipesilva: I agree with you there, but I'm not sure what a better name would be. Schematics also does it similarly with the `schematics` property in `collection.json`.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,262 @@
...
+    // Resolve root for the target.
+    // TODO: add Path format to JSON schemas
+    const target: Target<OptionsT> = {
+      root: resolve(this._root, normalize(project.root)),
+      projectType: project.projectType,
+      builder: projectTarget.builder,
+      options: {
+        ...options,
+        ...configuration ? configuration : {},
+        ...additionalInput ? additionalInput as {} : {},
```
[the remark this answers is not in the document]
filipesilva: Fixed.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,262 @@
...
+export interface Target<OptionsT = {}> {
+  root: Path;
+  projectType: string;
+  builder: string;
+  options: OptionsT;
+}
+
+export class Architect {
+  private _workspaceSchema = pathResolve(__dirname, 'workspace-schema.json');
+  private _buildersSchema = pathResolve(__dirname, 'builders-schema.json');
```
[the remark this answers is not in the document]
filipesilva: Done, I went the readonly route. Did not make them public because right now I don't see a reason for them to be exposed.

packages/angular_devkit/architect/src/architect.ts
filipesilva: (github got a bit confused and this comment is a duplicate)

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -0,0 +1,270 @@
...
+  builder: string;
+  options: OptionsT;
+}
+
+export class Architect {
+  private _workspaceSchema = join(normalize(__dirname), 'workspace-schema.json');
+  private _buildersSchema = join(normalize(__dirname), 'builders-schema.json');
+  private _workspace: Workspace;
+
+  constructor(private _root: Path, private _host: virtualFs.Host<{}>) { }
```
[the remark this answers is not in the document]
filipesilva: (github got a bit confused and this comment is a duplicate)

packages/angular_devkit/architect/src/architect.ts
filipesilva: (github got a bit confused and this comment is a duplicate)

packages/angular_devkit/architect/src/builder.ts
filipesilva: Yes, much better 👍

packages/angular_devkit/build_webpack/src/browser/index.ts
```diff
@@ -0,0 +1,258 @@
...
+  statsErrorsToString,
+  statsToString,
+  statsWarningsToString,
+} from '../angular-cli-files/utilities/stats';
+const webpackMerge = require('webpack-merge');
+
+
+// TODO: Use quicktype to build our TypeScript interfaces from the JSON Schema itself, in
+// the build system.
+export interface BrowserBuilderOptions {
```
clydin: Should we attempt standardize a minimal set of _browser_ options that all browser builds should support (or indicate that they support)?  This should ideally help with transitioning between different builders.

Or maybe some type of feature flagging capability that allows for a standard way to configure certain options (or option sets)?  Prime possibilities could be assets, external scripts, or i18n.

packages/angular_devkit/architect/src/workspace-schema.json
```diff
@@ -0,0 +1,106 @@
...
+      "additionalProperties": {
+        "$ref": "#/definitions/project"
+      }
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "name",
+    "version",
+    "root",
```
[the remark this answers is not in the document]
filipesilva: I'm actually not too sure. I have gone back and forth a bit with this one. Just using the file location leads to a bit of repetition in the project roots. But I think the default should be just where it is (`./`) regardless. I'll add that.

packages/angular_devkit/architect/src/workspace-schema.json
```diff
@@ -0,0 +1,106 @@
...
+    },
+    "root": {
+      "type": "string",
+      "description": "Workspace root."
+    },
+    "defaultProject": {
+      "type": "string",
+      "description": "Default target to run."
+    },
+    "projects": {
```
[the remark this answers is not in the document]
filipesilva: I would like that but it might be a moot point with working schema `$refs`. As is it's slightly out of scope of the initial implementation, but definitely an important enhancement for the future.

packages/angular_devkit/build_webpack/src/browser/index.ts
filipesilva: I tried to follow that train of thought for a while a month of so ago and didn't think we could hope to do it. Every setup uses different names for everything. And it wouldn't make much of a difference overall, because browser builds are just one of the many we need to do.

clydin: Well underneath the names could be different, but I think a big selling point of a facade is that it hides differences between the underlying implementations (i.e., standardized nomenclature and concepts on the user facing side with adapters and transformations on the inside).

packages/angular_devkit/architect/src/workspace-schema.json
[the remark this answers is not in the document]
hansl: Out of scope for now. My `pre`/`post` transform PR made it trivial to implement, so I'm fine with this PR as is.

packages/angular_devkit/architect/src/architect.ts
hansl: `targets`?

packages/angular_devkit/architect/src/architect.ts
hansl: Yes. `NullLogger` are preferred IMO. Same with Schematics.

packages/angular_devkit/architect/src/architect.ts
filipesilva: Please not another meaning for the word target :(

packages/angular_devkit/architect/src/architect.ts
filipesilva: Done.

hansl:
Merging this. Filipe has signed the CLA and the bot is just confused.

hansl:
Congrats to Filipe and everyone who worked hard on this. This is a great step!

PR #379, opened 2018-01-12
remove deprecated typeof-compare  from tslint.json

Sibiraj-S:
I hope this is the file that is used as `tslint.json` while creating new projects with angular-cli

The `typeof-compare` has been deprecated/redundant since `tslint 5.9.1`. Hence removing it

tslint v5.9.1 Changelog: https://github.com/palantir/tslint/releases/tag/5.9.1


markgoho-EDT:
Did this get reverted somewhere along the way? New angular cli projects are including it in the tslint

PR #384, opened 2018-01-13
fix(@angular-devkit/schematics): support existing CLI versions

clydin:


gareththomasnz:
thanks for the fast fix

PR #390, opened 2018-01-13
refactor(@angular-devkit/core): expose string utils as a scoped namespace

caroso1222:
This PR exposes the strings utils in `@angular-devkit/core` as a scoped namespace and refactors all the schematics that used the `strings` file inside `schematics/angular`. 

Note: This is a follow up of this PR: https://github.com/angular/devkit/pull/340. I messed up by rebasing that one so I'm opening a new one here.


filipesilva:
Thanks for refactoring this 👍

Issue #311, opened 2017-12-04
A flag to display a list of available parameters


sebawita:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Area
```
- [ ] devkit
- [x] schematics
```

### Desired functionality
Currently the only way to check the available parameters for a schematics is by looking into its json schema file.
It would be great to have a way in the CLI to display the list of params, like this:
```schematics @schematics:schematic --schema```

ssilvert:
+1

It seems like the proper flag for this would be --help.  Of course, --help is used for something else now, but perhaps schema help should be incorporated into that.

alexeagle:
This issue was moved to angular/angular-cli#12152

PR #391, opened 2018-01-14
feat(@angular-devkit/schematics-cli): add flag to print the schematic schema

caroso1222:
This PR adds support to print the schematic's schema using the proposed `--schema` flag. It's known that each schema property key can be named with any arbitrary string, but the print algorithm will just take some of the most known: "type", "default", "required", "description" and "alias". All keys different from those will not be printed, only the property name.

I tried to follow the same coloring format from `ng help`.

This is how it looks rendered in the console:

screen shot 2018-01-13 at 10 34 03 pm

I wanted to print "enums" as well as it seems like it's important, but I still can't wrap my head around how to render that. Any suggestions welcome.

Note: I didn't build any unit test because I didn't find any for flags like `--list-schematics` or `--help` so I'm not really sure if you are contemplating unit testing on these things or not. Happy to do it, just need some direction.

Closes https://github.com/angular/devkit/issues/311

### DIFF
--- a/packages/angular_devkit/schematics_cli/bin/schematics.ts
+++ b/packages/angular_devkit/schematics_cli/bin/schematics.ts
@@ -102,28 +102,28 @@ function parseSchematicName(str: string | null): { collection: string, schematic
 
 /**
  * Displays descriptions for each one of the properties inside a Schematic's Schema.
- * 
+ *
  * The print format is a follows:
  *   --<%= propName %> (<%= type %>): (Default: <%= default %>) (<%= required %>) <%= description %>
  *     alias: -<%= alias =>
- * 
+ *
  * A property is not printed if "visible":"false"
- * 
+ *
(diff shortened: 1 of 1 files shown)


filipesilva:
@hansl WDYT? Something like this would be useful for Build Facade as well. I'm unsure how this plays with `$ref` and more complicated defaults though.

hansl:
I think the flag should be `--help` instead.

hansl:
Otherwise this sounds like a good idea :)

alexeagle:
Sorry @caroso1222  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

caroso1222:
No problem @alexeagle, I understand. Thanks for the heads up.

PR #394, opened 2018-01-16
feat(@schematics/angular): upgrade to ng 5.2.0

akkumar:
ng 5.2.x released ~~and supports typescript 2.6.x as well~~. 

More details here at - 
https://blog.angular.io/angular-5-2-now-available-312d1099bd81 . 

Upgrade the generated package.json to refer to 5.2.x

### DIFF
--- a/packages/schematics/angular/application/files/package.json
+++ b/packages/schematics/angular/application/files/package.json
@@ -45,6 +45,6 @@
     "protractor": "~5.1.2",
     "ts-node": "~4.1.0",
     "tslint": "~5.9.1",<% } %>
-    "typescript": "~2.6.2"
+    "typescript": "~2.5.3"
   }
 }
--- a/packages/schematics/angular/application/files/package.json
+++ b/packages/schematics/angular/application/files/package.json
@@ -12,24 +12,24 @@
   },
   "private": true,
   "dependencies": {
-    "@angular/animations": "^5.1.0",
-    "@angular/common": "^5.1.0",
-    "@angular/compiler": "^5.1.0",
-    "@angular/core": "^5.1.0",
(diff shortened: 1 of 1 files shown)


akkumar:
```console
> concurrently "npm run serve" "npm run protractor" --kill-others --success first

sh: 1: concurrently: Permission denied
```

It is not obvious if the bug is related to this commit though.

filipesilva:
@akkumar that was a flake in the CI and is not related to your PR. I have restarted that CI job. Thanks for updating these dependencies!

clydin:
This is blocked on https://github.com/angular/angular-cli/pull/9097

cyrilletuzi:
We should not update to TS 2.6 yet, there is an important issue with the language service which is not working anymore in TS 2.6 : angular/angular#21591

akkumar:
Thanks @cyrilletuzi for bringing this one up. 

For now - I have reverted the ts 2.6 change back to what it was ( 2.5.3 ) to avoid issues. 

ng has been upgraded to 5.2.x though (with ts at 2.5.3 as before ).  Because it is back to 2.5.3 it should also not be dependent anymore on  angular/angular-cli#9097 - the issue that @clydin had referred to as well. 

@filipesilva @Brocco - can you help check this one and get it in as ng 5.2.x would still be compatible with 2.5.x , I assume ? 

Let me know your comments on the same.

cyrilletuzi:
For info, even if angular/angular#21591 is not marked as fixed, I don't have anymore problems with TypeScript 2.6 and VS Code Language Service. I think the extension patched the issue inside itself, so it should be safe to upgrade to TS 2.6.

PR #395, opened 2018-01-17
 feat(@schematics/angular): use tasks to install/init applications

clydin:


packages/angular_devkit/schematics/testing/schematic-test-runner.ts
```diff
@@ -58,6 +59,9 @@ export class SchematicTestRunner {
     const registry = new schema.CoreSchemaRegistry(formats.standardFormats);
 
     this._engineHost.registerOptionsTransform(validateOptionsWithSchema(registry));
+    this._engineHost.registerTaskExecutor(BuiltinTaskExecutor.NodePackage);
```
hansl: There should be a way to test that a task is added, without running the test. TestRunner never actually commit.

PR #398, opened 2018-01-19
feat(@angular-devkit/schematics): support collection extension

### LINKED ISSUE #34: Collection Extension / Merging
Sometimes you want to reuse schematics from another collection as is. For example, someone could want to implement a collection that is the exact same as the angular one (same schematics), but changes only 1 schematic. All the others should be the same. Or maybe a company wants to have a default collection on their CLI but have that collection include schematics from a number of other collections.

A proper system should allow merging of collections, with some resolution if there's a collusion, and a list of schematics that it implements itself (might be empty). As an example (the schema is just an example):

```json
{
  "$schema": "http://schema.angular.io/schematics/collection/2",
  "name": "my-com",
  "description": "My Company default collection, containing all schematics that are used normally",
  "extends": [
    "@schematics/angular",
    "@my-company/schematics-base",
    "@my-company/schematics-pages"
  ],
  "schematics": {
  }
}
```

This would include the base angular schematics, the base my-company schematics and the pages schematics. This helps with separating development of each of those, while the end user can just install and use one collection. The schematics are merged, with names being overridden by pages, then base, then angular.

clydin:
Example usage (also supports an array of strings):
```{
  "name": "extends-basic-string",
  "extends": "works",
  "schematics": {
    "schematic2": {
      "description": "2",
      "factory": "../null-factory"
    }
  }
}
```

Closes #34

### DIFF
--- a/packages/angular_devkit/schematics/collection-schema.json
+++ b/packages/angular_devkit/schematics/collection-schema.json
@@ -4,6 +4,22 @@
   "title": "Collection Schema for validating a 'collection.json'.",
   "type": "object",
   "properties": {
+    "extends": {
+      "oneOf": [
+        {
+          "type": "string",
+          "minLength": 1
+        },
+        {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "minLength": 1
+          },
+          "minItems": 1
+        }
+      ]
+    },
     "schematics": {
       "type": "object",
       "description": "A map of schematic names to schematic details",
--- a/packages/angular_devkit/schematics/src/engine/engine.ts
+++ b/packages/angular_devkit/schematics/src/engine/engine.ts
@@ -40,6 +40,13 @@ export class UnknownUrlSourceProtocol extends BaseException {
 export class UnknownCollectionException extends BaseException {
   constructor(name: string) { super(`Unknown collection "${name}".`); }
 }
+
+export class CircularCollectionException extends BaseException {
+  constructor(name: string) {
+    super(`Circular collection reference "${name}".`);
+  }
+}
+
 export class UnknownSchematicException extends BaseException {
   constructor(name: string, collection: CollectionDescription<{}>) {
     super(`Schematic "${name}" not found in collection "${collection.name}".`);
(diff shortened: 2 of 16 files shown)


filipesilva:
@clydin can you rebase please, and @hansl can you review?

packages/angular_devkit/schematics/collection-schema.json
```diff
@@ -4,6 +4,11 @@
   "title": "Collection Schema for validating a 'collection.json'.",
   "type": "object",
   "properties": {
+    "extends": {
```
hansl: How hard to add `type: [ "array", "string" ]` ?

packages/angular_devkit/schematics/tools/file-system-engine-host_spec.ts
```diff
@@ -38,6 +38,66 @@ describe('FileSystemEngineHost', () => {
     expect(schematic1.description.name).toBe('schematic1');
   });
 
+  it('extends a collection ', () => {
```
hansl: Should just add a test for `listSchematicNames` to avoid regression.

packages/angular_devkit/schematics/collection-schema.json
clydin: Added support for both.

packages/angular_devkit/schematics/tools/file-system-engine-host_spec.ts
clydin: done.

packages/angular_devkit/schematics/src/engine/engine.ts
```diff
@@ -76,16 +83,38 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
...
+    collection = new CollectionImpl<CollectionT, SchematicT>(description, this, bases);
+    this._collectionCache.set(name, collection);
+    this._schematicCache.set(name, new Map());
+
+    return collection;
+  }
+
+  private _createCollectionDescription(
+    name: string,
+    parentNames?: Set<string>,
```
hansl: Just one nit but not stopping the approval; `parentNames = new Set<string>()` then lower just add regardless of whether it's defined or not.

PR #400, opened 2018-01-21
feat(@schematics/angular): add flag for IE/Edge to prevent error or performance impact

JiaLiPassion:
from `zone.js` 0.8.13, there is a new flag `__Zone_enable_cross_context_check` was added.

- IE/Edge: in Developer tools, `addEventListener` will also be patched by `zone.js`. 
 -  in Edge the performance will be slowed down, and the `performance monitor` data will be wrong because of the `zone.js` patch.
 - in some version of `IE11`, the `addEventListener` will throw exception because of the `security` error.

`zone.js` will check those cases and if the code is run in `IE/Edge` developer tools, just use `native addEventListener` to prevent such kind of performance impact or error.

To do this check, `zone.js` will call `callback.toString()` (callback is the handler in  `addEventListener(eventName, callback);`) , and in `IE/Edge` and some version of `Firefox`, the `callback.toString()` will throw error because of `security` reason.

and the purpose of the `__Zone_enable_cross_context_check` flag is 
- when the flag is true, `zone.js` will catch the `callback.toString()` possible error, and use `native addEventListener`

so in this PR, I add the flag in `polyfill.ts`, user can set the flag to true when user want to debug in `IE/Edge` when development and can set the flag to false in production mode.

I don't know is that possible to know the `dev/prod` mode in `polyfill.ts`, if we can, we can check the mode and only set the flag to true when we are in dev mode.

### DIFF
--- a/packages/schematics/angular/application/files/__sourcedir__/polyfills.ts
+++ b/packages/schematics/angular/application/files/__sourcedir__/polyfills.ts
@@ -61,6 +61,12 @@ import 'core-js/es7/reflect';
  // (window as any).__Zone_disable_on_property = true; // disable patch onProperty such as onclick
  // (window as any).__zone_symbol__BLACK_LISTED_EVENTS = ['scroll', 'mousemove']; // disable patch specified eventNames
 
+ /*
+ * in IE/Edge developer tools, the addEventListener will also be wrapped by zone.js
+ * with the following flag, it will bypass `zone.js` patch for IE/Edge
+ */
+// (window as any).__Zone_enable_cross_context_check = true;
+
 /***************************************************************************************************
  * Zone JS is required by default for Angular itself.
  */


filipesilva:
@JiaLiPassion can you give us some more information about these changes please:
- does not having this flag set affect users visiting a deployed app in IE/Edge/Firefox?
- which versions of Edge/Firefox are affected?
- in what version of zone.js were these changes released?

@hansl PTAL as well.

JiaLiPassion:
@filipesilva , thank you for review.

> does not having this flag set affect users visiting a deployed app in IE/Edge/Firefox?
- for functionality, there is no affect. but with this flag on, there is performance impact when `addEventListener` was called.

> which versions of Edge/Firefox are affected?

from the issues users posted in `zone.js` repo,
- IE9
- IE10
- IE11
- Edge 14
- Edge 15
- Firefox 52

was affected.

> in what version of zone.js were these changes released?

this flag was released in zone.js `0.8.13`.

filipesilva:
Just to clarify, you mentioned
> in IE/Edge and some version of Firefox, the callback.toString() will throw error because of security reason.

Won't this error crash the app if a developer does not set the flag?

JiaLiPassion:
@filipesilva , I am not sure the exact conditions to reproduce this issue.
without this flag, sometimes error will thrown, such as 
- https://github.com/angular/zone.js/issues/933
- https://github.com/angular/zone.js/issues/950

it may have some connections with user's security settings.

filipesilva:
That looks worrying. It sounds to me like that flag should be always true, otherwise user apps can break in hard to debug ways with the new zone.js. 

According to https://angular.io/guide/browser-support, Edge 14 is still supported, and so are the IE versions listed.

Current apps that are using an loose version of zone.js might already be breaking in production.

/cc @IgorMinar

JiaLiPassion:
@filipesilva , yes, you are right, for the issues I met, they are all related to `IE/Edge developer tools`.
so if user open `developer tools` in production, it may have the same issue.

mhevery:
My understand is that this is an error which would occur only when opening Developer Tools on IE, so having the default this way will make it fine for users, but may effect some developers. Given that I think perf for user beats inconveniencing few developers on IE.

filipesilva:
If it only happens on the developer tools then it's not a problem for deployed apps. According to https://angular.io/guide/browser-support only latest Firefox (58) is supported so it should be ok as well if the issue only affected Firefox 52.

Thanks @JiaLiPassion and @mhevery for all the details.

PR #413, opened 2018-01-26
style: Add tslint rule to prevent using non-pipeable imports

Brocco:
This should be part of the next minor release, not a patch release as it could cause issues with consumers of the schematics library that are not using pipeable operators.


clydin:
Bump the version here as well (just in case)? https://github.com/angular/devkit/blob/master/.monorepo.json#L78

hansl:
@clydin that's updated automatically with `devkit-admin release`. Shouldn't need to update that line.

PR #416, opened 2018-01-30
Add Bazel build files for build_optimizer, and a rollup plugin wrapper

alexeagle:


filipesilva:
@alexeagle I think you have a bad rebase here. Your commit also needs to use the commit format.

alexeagle:
This is on hold, until I get it working on the other end.

packages/angular_devkit/build_optimizer/src/build-optimizer/rollup-plugin.ts
```diff
@@ -0,0 +1,37 @@
...
+import { buildOptimizer } from './build-optimizer';
+import { RawSourceMap } from 'source-map';
+
+const DEBUG = false;
+
+export default function optimizer() {
+  return {
+    name: 'build-optimizer',
+    transform: (content: string, id: string): {code: string, map: RawSourceMap}|null => {
+      const isSideEffectFree = id.indexOf("packages/core/core.es5_esm") >= 0;
```
filipesilva: Can you make this a config item for the plugin instead of a hardcoded path please?

alexeagle:
Works in concert with usage in Angular repo now, PTAL

packages/angular_devkit/build_optimizer/src/build-optimizer/rollup-plugin.ts
alexeagle: done

filipesilva:
Changes LGTM but commit message needs to be something like `feat(@angular-devkit/build-optimizer): add rollup plugin & bazel build file`

filipesilva:
I can confirm @alexeagle has signed the CLA and consents to his code being used.

PR #421, opened 2018-02-01
refactor: added a tslint rule to enforce package.json dependencies

hansl:
And added disable of the rule to spec and scripts files.

### DIFF
--- a/packages/angular_devkit/core/src/virtual-fs/path.ts
+++ b/packages/angular_devkit/core/src/virtual-fs/path.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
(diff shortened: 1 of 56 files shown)


packages/angular_devkit/core/src/virtual-fs/path.ts
```diff
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { BaseException } from '@angular-devkit/core';
+import { BaseException } from '..';
```
clydin: I like the idea of the rule but forcing a change such as this, I think this causes a decrease in readability and overall code quality.  The rule should at a minimum allow the name of the package itself and should ideally support any path mapping.  The second part especially as the rule pretty much makes advanced path mapping unusable.

PR #422, opened 2018-02-03
fix(main): match styleguide catch callback

seangwright:
https://angular.io/guide/styleguide#style-02-05

In the styleguide rule above the `catch` callback uses `console.error` when the app boostrapping fails. I am updating the template to match.

I wasn't sure if the success `console.log` should also be included but that seems a bit too opinionated.


alexeagle:
Sorry @sgwatgit  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #423, opened 2018-02-05
fix(@angular-devkit/schematics): VirtualDirEntry#subdirs & proper testing

ymeine:
- the getter `subdirs` in class `VirtualDirEntry` was wrongly implemented,
returning irrelevant data. Make some code common between `subdirs`
and `subfiles` for clarity and DRY respect.
- testing was not done properly either, since a `FileSystemTree` instance was used
(creating corresponding `FileSystemDirEntry` instances) in a test for `VirtualTree`
(and its corresponding model `VirtualDirEntry`). Despite the inheritance,
it wouldn't test it properly. Therefore copy this test as is to the proper file
`filesystem_spec.ts` and slightly adapt it in `virtual_spec.ts` to use the proper
implementations instead.

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/virtual.ts
+++ b/packages/angular_devkit/schematics/src/tree/virtual.ts
@@ -49,15 +49,27 @@ export class VirtualDirEntry implements DirEntry {
     return this._path == '/' ? null : this._tree.getDir(dirname(this._path));
   }
   get path() { return this._path; }
+
   get subdirs() {
-    return this._tree.files
-               .filter(path => path.startsWith(this._path) && dirname(path) != this._path)
-               .map(path => basename(path));
+    const directChildPartsCount = split(normalize(this._path)).length + 1;
+
+    const directories = this._tree.files
+    // make sure entries belong to proper subbranch
+    .filter(path => path.startsWith(this._path))
+    // get all existing directories & prune to direct children
+    .map(path => split(normalize(path)).slice(0, -1).slice(0, directChildPartsCount))
+    // exclude current directory
+    .filter(parts => parts.length === directChildPartsCount)
+    // get directory name only
+    .map(parts => parts[parts.length - 1]);
+
+    // make sure to have a unique set (directories contain multiple files so appear multiple times)
+    return Array.from(new Set(directories));
   }
   get subfiles() {
     return this._tree.files
-               .filter(path => path.startsWith(this._path) && dirname(path) == this._path)
-               .map(path => basename(path));
+      .filter(path => dirname(path) === this._path)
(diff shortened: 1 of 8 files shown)


ymeine:
@googlebot

packages/angular_devkit/schematics/src/tree/virtual.ts
```diff
@@ -49,15 +49,20 @@ export class VirtualDirEntry implements DirEntry {
     return this._path == '/' ? null : this._tree.getDir(dirname(this._path));
   }
   get path() { return this._path; }
+
+  private _getDirectChildren(paths: Path[]) {
+    return paths
+      .filter(path => dirname(path) === this._path)
+      .map(path => basename(path));
```
hansl: This isn't proper logic. Your DRY has flaws in it.

If `files` is the following:

```js
[
  '/sub1/sub2/sub3/file1',
  '/sub1/sub4/sub5/file2',
  '/sub1/file3',
]
```

Listing `/sub1` should result in `[sub2, sub4, file3]`, but it won't using that new logic (it will list `[file1, file2, file3]`). There is a reason it wasn't DRY; the logic is different when listing files and dirs.

hansl:
I think this PR is on the right track to fix the issue. It just needs to use prefixes instead of `dirname()`. Remember the `files` array is just an array of all file paths. So directories can jump from 1 sub to n sub.

packages/angular_devkit/schematics/src/tree/virtual.ts
ymeine: Indeed, after diving into the code for the first and investigating hard, I didn't have sufficient resources anymore to design it properly I guess... I'm working on the fix right now :) (and of course on enhancing the current test cases accordingly)

ymeine:
Fixing it properly was not that trivial finally, and led to more fixes and more test cases.

I added commits instead of squashing them for now, so you can review the changes only if you wish.

I added two different commits, since the last one refactors a bit the testing code I added but I'm not sure it suits your coding conventions. Not sure either for the comments I left in the code, even though I think they truly help understanding. Let me know what you think.

hansl:
Hi @ymeine,

We reworked the schematics hosts to use the new Virtual FS. Could you rebase on top of it?

ymeine:
@hansl Done, you might want to review the changes again even if it didn't change much. To summarize: 

- now when trying to visit a `FileSystemTree` instance given a path to a file an exception is thrown, checking that in the test
- the fix I did for `InMemoryFileSystemTreeHost` went away with the file, but I checked the new implementation in `SimpleMemoryHost#_isDir` and I added support for the edge case of the root path `/` (with according test)

PR #431, opened 2018-02-08
reorder package.json scripts

Sibiraj-S:
Just sorted alphabetically.

Reason,
When there is a increase in number of scripts. Alphabetically arranging is little easing.

### DIFF
--- a/packages/schematics/angular/application/files/package.json
+++ b/packages/schematics/angular/application/files/package.json
@@ -3,12 +3,12 @@
   "version": "0.0.0",
   "license": "MIT",
   "scripts": {
+    "e2e": "ng e2e",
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/application/files/package.json
```diff
@@ -3,12 +3,12 @@
   "version": "0.0.0",
   "license": "MIT",
   "scripts": {
-    "ng": "ng",
-    "start": "ng serve",
+    "e2e": "ng e2e",
```
hansl: Wrong order with `build`.

Sibiraj-S: @hansl . Since I deleted the fork. Made another PR with these changes #458

PR #432, opened 2018-02-09
JsonSchema QoL changes

hansl:


packages/angular_devkit/core/src/json/schema/interface.ts
```diff
@@ -6,9 +6,11 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import { Observable } from 'rxjs/Observable';
+import { JsonValue } from '..';
 
 
 export interface SchemaValidatorResult {
+  data: JsonValue;
```
filipesilva: Does having the data be returned here means the original values are no longer altered? That would be useful overall. It's kinda unintuitive that the original is mutated now.

packages/angular_devkit/core/src/json/schema/visitor.ts
```diff
@@ -90,6 +110,18 @@ function _visitJsonRecursive(
               value[key],
               visitor,
               joinJsonPointer(ptr, key),
+              (typeof schema == 'object'
```
filipesilva: This really looks like it should be in a helper function with some explanation of what the purpose is.

filipesilva: This really looks like it should be in a helper function with some explanation of what the purpose is.

packages/angular_devkit/core/src/json/schema/visitor.ts
```diff
@@ -110,13 +142,18 @@ function _visitJsonRecursive(
  *
  * @param {JsonValue} json The Json value to visit.
  * @param {JsonVisitor} visitor A function that will be called on every items.
+ * @param {JsonObject} schema A JSON schema to pass through to the visitor (where possible).
+ * @param refResolver a function to resolve references in the schema.
  * @returns {Observable< | undefined>} The observable of the new root, if the root changed.
  */
-export function visitJson(
+export function visitJson<ContextT>(
```
filipesilva: Is there any guarantee that the JsonValue is not cyclical? If you want to accept recursive JsonValues, you'll need to transform this into a depth first with open/closed lists to not be stuck in loops.

Some of the AJV operations can end up having references instead of values.

hansl: There's none. We cannot really go through that either (it would be too memory and time consuming). But at that point isn't it the same as an infinite loop on the part of the API user?

packages/angular_devkit/core/src/json/schema/visitor.ts
hansl: Done.

packages/angular_devkit/core/src/json/schema/interface.ts
hansl: You get a visit for the initial value, so if you return a different value then we cannot mutate the original data.

Making a copy of the data is impossible at that point, because some of it might not be JSON values but objects and whatnot.

packages/angular_devkit/core/src/json/schema/visitor.ts
```diff
@@ -0,0 +1,184 @@
...
+    pointer: JsonPointer,
+    root?: JsonObject | JsonArray,
+  ): Observable<JsonValue | undefined> | JsonValue | undefined;
+}
+
+
+export function buildJsonPointer(fragments: string[]): JsonPointer {
+  return (
+    '/' + fragments.map(f => {
+      return f.replace(/~/g, '~0')
```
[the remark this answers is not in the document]
hansl: `buildJsonPointer` takes a list of fragments and return the JSON pointer. What you're talking about is `parseJsonPointer` which does what you're describing.

Calling `parseJsonPointer('/~001')` would return `['~01']` which is the correct value you passed to `buildJsonPointer`.

hansl:
@filipesilva All done. Please take another look.

PR #436, opened 2018-02-13
feat(@angular-devkit/schematics): use the Virtual FS Host in Schematics

hansl:


packages/angular_devkit/schematics/tools/file-system-host.ts
```diff
@@ -5,28 +5,11 @@
...
-  readFile(path: string) {
-    return readFileSync(join(this._root, path));
-  }
-  exists(path: string) {
-    return existsSync(this.join(this._root, path));
-  }
-
-  join(path1: string, path2: string) {
-    return join(path1, path2);
+export class FileSystemHost extends virtualFs.DelegateHost<{}> {
```
clydin: Could this be implemented as:
```
export class FileSystemHost extends virtualFs.ScopedHost {
    constructor(dir: string) {
        super(new NodeJsSyncHost(), normalize(dir));
    }
 }
```

Seems to be the only place the `DelegateHost` class is used.

hansl: Yes, probably. Done.

packages/angular_devkit/core/src/virtual-fs/host/test.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(map: { [path: string]: string } = {}) {
+    super();
+
+    for (const filePath of Object.getOwnPropertyNames(map)) {
+      this.write(normalize(filePath), stringToFileBuffer(map[filePath]));
+    }
+    debugger;
+  }
+
+  get files() { return [...this._cache.keys()]; }
```
clydin: Thoughts on moving this (`files`) up into `SimpleMemoryHost`?  Wouldn't need to expose the underlining cache object then.

PR #444, opened 2018-02-18
feat: build architect for ng-packagr

dherges:
Schematic for library scaffolding:

- `ng new library foo` will 
   - create folder `lib/foo/package.json` with ngPackage config
   - update tsconfig for dev experience
   - update root package.json with additional dependencies

Build architect for ng-packagr:
 - forward project param from ng-packagr


packages/angular_devkit/build_ng_packagr/src/build/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    return new Observable(obs => {
+      const projectNgPackagr = requireProjectModule(root, 'ng-packagr') as typeof ngPackagr;
+      const packageJsonPath = resolvePath(root, options.project);
+
+      projectNgPackagr.ngPackagr()
+        .forProject(packageJsonPath)
+        .build()
+        .then(() => obs.complete())
+        .catch((e: any) => obs.error(e));
```
dherges: At the moment, the only way is passing a file path in either [`ngPackagr.forProject()`](https://github.com/dherges/ng-packagr/blob/master/src/lib/ng-v5/packagr.ts#L28-L38) or in the [`build()` command](https://github.com/dherges/ng-packagr/blob/master/src/lib/commands/build.command.ts#L4-L22)

dherges:
Note: windows test failures due to path delimeters. https://ci.appveyor.com/project/AngularCLI/devkit/build/1.0.740#L82

packages/schematics/angular/library/index.ts
```diff
@@ -0,0 +1,111 @@
...
+        'typescript': '>=2.4.2',
+        // de-structure last keeps existing user dependencies
+        ...json.devDependencies
+      };
+
+      // Add package.json script
+      if (!json.scripts) {
+        json.scripts = {};
+      }
+      json.scripts[`libs:${name}:build`] = `ng-packagr -p ${sourceDir}/package.json`;
```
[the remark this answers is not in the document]
filipesilva: I think the `ng` command to build is the one that should be called here, which in turn would call architect with the workspace setup. 

Although I do wonder if we should be adding the individual scripts at all when adding a new project/lib via schematic.

packages/angular_devkit/build_ng_packagr/src/build/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  run(target: Target<NgPackagrBuilderOptions>): Observable<BuildEvent> {
+    const root = getSystemPath(target.root);
+    const options = target.options;
+
+    if (!options.project) {
+      throw new Error('A "project" must be specified to build a library\'s npm package.');
+    }
+
+    return new Observable(obs => {
+      const projectNgPackagr = requireProjectModule(root, 'ng-packagr') as typeof ngPackagr;
```
filipesilva: I'm actually not sure either NgPackager, nor TsLint (where this setup code came from) should be loaded as project dependencies.

I think it should either be a peer dependency, or a direct dependency. Both of those approaches have pros and cons, but in either case I don't think we should try to short-circuit it by requiring the project module instead of importing directly.

In the case of NgPackagr, you already have it as a direct dependency so I think we can import it directly.

packages/angular_devkit/build_ng_packagr/src/build/index_spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      },
+    },
+  });
+
+  it('runs', (done) => {
+    const architect = new Architect(normalize(root), host);
+    architect.loadWorkspaceFromJson(getWorkspace()).pipe(
+      concatMap((architect) => architect.run(architect.getTarget())),
+      toArray(),
+      tap(events => expect(events.length).toBe(0)),
```
filipesilva: At the when I linked you the TSlint example the tests there were a bit anemic. Since then I added more, and more involved ones.

I think here you should check the build event success, and also check if the output is what you expected. An example of a test that does that is https://github.com/angular/devkit/blob/master/packages/angular_devkit/build_webpack/test/tslint/works_spec_large.ts#L54-L67.

packages/angular_devkit/build_ng_packagr/src/build/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      const packageJsonPath = resolvePath(root, options.project);
+
+      projectNgPackagr.ngPackagr()
+        .forProject(packageJsonPath)
+        .build()
+        .then(() => obs.complete())
```
filipesilva: Overall we're trying to go for a model where builders always emit events. This wasn't the case before when you started working on it but since then we revised that because otherwise it was harder to test and reason about it.

The general idea is:
- whenever a builder finished a build, it will emit `{success: true}` if it succeeded or `{success: false}` if it did not
- when a builder is done, it will complete. Builders on watch mode would emit multiple times and not complete, whereas single run builders should emit once and complete.
- when there is an error, the builder should `obs.error`. Note that this is different from a failed build though, which emits a `{success: false}` and no error. I think the usage of error you have here is the right one.

tests/@angular_devkit/build_ng_packagr/ng-packaged/.angular-cli.json
```diff
@@ -0,0 +1,60 @@
+{
```
filipesilva: I know these files are needed for now but we need to review the final file set for adding a library to an existing project before wrapping up.

packages/schematics/angular/library/index.ts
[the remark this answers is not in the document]
hansl: Can we just hook past your CLI? You should probably expose a JavaScript API that we can use directly.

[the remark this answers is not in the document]
dherges: Yes, the schematic should create the workspace setup for Builder/Runner/Architect. What is the command users will execute then?

tests/@angular_devkit/build_ng_packagr/ng-packaged/.angular-cli.json
dherges: Should I strip the test to be a minimal sample? WOuld only be a `package.json`, a few `*.ts`/`*.html`/`*.css` files and then I'll verify the `dist` folder is created.

hansl:
Validating David's CLA, signed on Mar 31, 2017 02:43 PDT.

tests/@angular_devkit/build_ng_packagr/ng-packaged/.angular-cli.json
filipesilva: Yes that would be preferable, that way the tests only include what they need and it is easier to keep them updated.

filipesilva:
I followed up the comments in this PR in https://github.com/angular/devkit/pull/532.

PR #446, opened 2018-02-19
fix(@angular-devkit/architect-cli): fix package location

filipesilva:


akkumar:
Reverting #445 seemed to get rid of the unrelated build failure here. Similar issue in #448 . FYI .

filipesilva:
@akkumar correct, it is now fixed in master.

PR #449, opened 2018-02-20
build: compile with TS 2.7

clydin:
This adds some temporary patches for rxjs.  Once v6.0 is released and the devkit transitions, we can remove them.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,7 +22,7 @@
     // source in devtools.
     "inlineSources": true,
     "strictNullChecks": true,
-    "target": "es6",
+    "target": "es2016",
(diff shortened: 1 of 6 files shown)


tsconfig.json
```diff
@@ -56,6 +56,7 @@
     "node_modules/**/*",
     "packages/schematics/*/*/*files/**/*",
     "tmp/**/*",
+    "scripts/**/*",
```
hansl: `scripts/patches/**/*`. Seems like we want to ignore patches, but still compile and lint scripts.

clydin: Done.

PR #458, opened 2018-02-22
reorder package.json scripts alphabetically

Sibiraj-S:
Duplicate of #431

### DIFF
--- a/packages/schematics/angular/application/files/package.json
+++ b/packages/schematics/angular/application/files/package.json
@@ -3,12 +3,12 @@
   "version": "0.0.0",
   "license": "MIT",
   "scripts": {
+    "build": "ng build --prod",
+    "e2e": "ng e2e",
+    "lint": "ng lint",
     "ng": "ng",
     "start": "ng serve",
-    "build": "ng build --prod",
-    "test": "ng test",
-    "lint": "ng lint",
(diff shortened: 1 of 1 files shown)


Sibiraj-S:
@filipesilva. Nothing important actually. 

When there are a number of scripts, sorting in any order or following something pattern will have an effect on user's accessibility.

In my case,
I usually have a lot of scripts. I felt having scripts sorted, eases the pain while reading and finding them.

With the initial `6` scripts. It looks great. But In my case, I reorder every time.

Sorting Alphabetically is just a proposal. 

`Patterns` are good when things get complex

filipesilva:
Heya @Sibiraj-S, we've been talking a bit about this over the last few days to try and figure out if it would be a good change.

I had a few people saying they prefer it as is, and some saying that they prefer alphabetical like your proposed change. I think at the end of the day is a matter of style. 

Across the CLI/DevKit repos we group by functionality, and we've had the package.json of the starter projects grouped this way as well for a while. so I think there isn't a real benefit to changing it at the moment. 

I'm sorry that we're not merging your change but I think it was worth bringing it up because otherwise it wouldn't have been discussed at all. Thank you.

Sibiraj-S:
Thanks @filipesilva and all for the time. 😄

PR #460, opened 2018-02-22
proposal: proposal for ng update

hansl:


hansl:
Closed in favor of #473.

PR #461, opened 2018-02-23
ci: allow release scopes in commit messages outside of CI

hansl:


filipesilva:
@hansl I cannot merge this PR since it requires a code owner review from you... And I don't think you can review your own PRs.

image

PR #462, opened 2018-02-24
feat(@schematics/angular): strict option

cyrilletuzi:
`--strict` option to configure TypeScript in strict mode when generating an app.

Replaces #397 

@hansl @filipesilva


cyrilletuzi:
@filipesilva @Brocco Any news of this ? @hanls approved the idea of this PR in #397.

lacolaco:
This is very helpful, but I'm concerned about its name. I think `strict` is a bit ambiguous and may tend to conflict. in fact, JavaScript has `use strict` pragma.

cyrilletuzi:
Redone in #462 as I completely messed with git.

@lacolaco The name was approved by @hansl in #397 because it could be use for future other strict options.

PR #465, opened 2018-02-25
feat(@angular-devkit/schematics): add a workflow class that simplifies the whole thing

hansl:
Workflows are complete abstraction of the whole system. This makes it easier to
have tools that dont diverge from the standard definition of schematics. In case
tools have to diverge they can implement their own workflow as well.

Please note that workflows are also recursive. This means that you can execute
a schematic inside another schematic. This will happen in the next commit with
the schematic task.

For now, the workflow is relatively crude. It will be improved over then next
few commits.

### DIFF
--- a/packages/angular_devkit/schematics/src/engine/engine.ts
+++ b/packages/angular_devkit/schematics/src/engine/engine.ts
@@ -13,6 +13,7 @@ import { Url } from 'url';
 import { MergeStrategy } from '../tree/interface';
 import { NullTree } from '../tree/null';
 import { empty } from '../tree/static';
+import { Workflow } from '../workflow';
 import { CollectionImpl } from './collection';
 import {
   Collection,
@@ -78,9 +79,10 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
     = new Map<string, Map<string, SchematicImpl<CollectionT, SchematicT>>>();
   private _taskSchedulers = new Array<TaskScheduler>();
 
-  constructor(private _host: EngineHost<CollectionT, SchematicT>) {
+  constructor(private _host: EngineHost<CollectionT, SchematicT>, protected _workflow?: Workflow) {
   }
 
+  get workflow() { return this._workflow || null; }
   get defaultMergeStrategy() { return this._host.defaultMergeStrategy || MergeStrategy.Default; }
 
   createCollection(name: string): Collection<CollectionT, SchematicT> {
--- a/packages/angular_devkit/schematics/tasks/run-schematic/init-task.ts
+++ b/packages/angular_devkit/schematics/tasks/run-schematic/init-task.ts
@@ -1,0 +1,29 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { TaskConfiguration, TaskConfigurationGenerator } from '../../src';
+import { RunSchematicName, RunSchematicTaskOptions } from './options';
+
+
+export class RunSchematicTask implements TaskConfigurationGenerator<RunSchematicTaskOptions> {
+  constructor(
+    protected _collection: string,
+    protected _schematic: string,
+    protected _options: object,
+  ) {}
+
+  toConfiguration(): TaskConfiguration<RunSchematicTaskOptions> {
+    return {
+      name: RunSchematicName,
+      options: {
+        collection: this._collection,
+        name: this._schematic,
+        options: this._options,
+      },
+    };
+  }
+}
--- a/packages/angular_devkit/schematics/tools/workflow/node-workflow.ts
+++ b/packages/angular_devkit/schematics/tools/workflow/node-workflow.ts
@@ -1,0 +1,137 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { logging, schema, virtualFs } from '@angular-devkit/core';
+import {
+  DryRunSink,
+  HostSink,
+  HostTree,
+  SchematicEngine,
+  Sink,
+  Tree,
+  UnsuccessfulWorkflowExecution,
+  formats,
+  workflow,
+} from '@angular-devkit/schematics';  // tslint:disable-line:no-implicit-dependencies
+import { Observable } from 'rxjs/Observable';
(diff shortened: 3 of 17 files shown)


packages/angular_devkit/schematics/tasks/run-schematic/init-task.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { TaskConfiguration, TaskConfigurationGenerator } from '../../src';
+import { RunSchematicName, RunSchematicTaskOptions } from './options';
+
+
+export class RunSchematicTask implements TaskConfigurationGenerator<RunSchematicTaskOptions> {
+  constructor(
+    protected _collection: string,
```
clydin: should this support the combined string format for the collection and schematic as well?

packages/angular_devkit/schematics/src/engine/engine.ts
```diff
@@ -78,9 +79,10 @@ export class SchematicEngine<CollectionT extends object, SchematicT extends obje
     = new Map<string, Map<string, SchematicImpl<CollectionT, SchematicT>>>();
   private _taskSchedulers = new Array<TaskScheduler>();
 
-  constructor(private _host: EngineHost<CollectionT, SchematicT>) {
+  constructor(private _host: EngineHost<CollectionT, SchematicT>, protected _workflow?: Workflow) {
   }
 
+  get workflow() { return this._workflow || null; }
```
clydin: Exposing this downward into the implementation interweaves the higher level abstraction of the `Workflow` with the rest of the system.  I think ideally a schematic or task shouldn't have to rely on the Workflow or that a Workflow even exists.

Some ideas on how this could be implemented in the comments for the node workflow.

packages/angular_devkit/schematics/tools/workflow/node-workflow.ts
```diff
@@ -0,0 +1,133 @@
...
+    const parentContext = this._context[this._context.length - 1];
+    const context = {
+      ...options,
+      debug: options.debug || false,
+      logger: options.logger || new logging.NullLogger(),
+      parentContext,
+    };
+    this._context.push(context);
+
+    return schematic.call(options.options, of(new HostTree(this._host)), {
```
clydin: What do you think about inverting the responsibility here and having the engine `execute` the schematic.  This would give the engine access to the host and even the sinks as they could be passed in as a  _ tree sink_ Observable parameter.  So the passed in sinks would be an observable chain that would get access to the event stream and then either act on them, drop them, pass them along, or some combination.

packages/angular_devkit/schematics/src/engine/task.ts
```diff
@@ -33,73 +27,38 @@ export interface TaskExecutorFactory<T> {
...
 
 export class TaskScheduler {
-  private _queue = new PriorityQueue<TaskInfo>((x, y) => x.priority - y.priority);
+  private _tasks = new PartiallyOrderedSet<TaskId>();
   private _taskIds = new Map<TaskId, TaskInfo>();
   private static _taskIdCounter = 1;
 
   constructor(private _context: SchematicContext) {}
 
-  private _calculatePriority(dependencies: Set<TaskInfo>): number {
```
clydin: I kind of liked the deceptive simplicity of the priority queue based on dependency depth.

I was considering moving it down one step in the pipeline and building a graph of all the tasks first so that they could be analyzed and potentially re-configured/arranged based on their dependencies and configuration.

packages/angular_devkit/schematics/tasks/run-schematic/init-task.ts
```diff
@@ -0,0 +1,29 @@
+/**
```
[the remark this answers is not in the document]
hansl: done. I'll change others too.

packages/angular_devkit/schematics/tasks/run-schematic/init-task.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+import { TaskConfiguration, TaskConfigurationGenerator } from '../../src';
+import { RunSchematicName, RunSchematicTaskOptions } from './options';
+
+
+export class RunSchematicTask implements TaskConfigurationGenerator<RunSchematicTaskOptions> {
+  constructor(
+    protected _collection: string,
+    protected _schematic: string,
+    protected _options: object,
```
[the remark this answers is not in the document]
hansl: You need to be able to pass in options as inputs to the new workflow.

packages/angular_devkit/schematics/tasks/run-schematic/init-task.ts
hansl: The goal of this task is to be created/ran from code. The `collection:schematic` syntax is just convention from command line.

packages/angular_devkit/schematics/src/engine/task.ts
hansl: I'm okay removing this commit. `PartiallyOrderedSet` is different than the priority queue in a lot of things, and the `TaskScheduler` also had a different behaviour too; it's not a set. Running the same configuration twice would actually run it twice. In the new code it wouldn't.

Just let me know if you want me to remove it and I'll do that. I started work in the TaskScheduler to allow for schematic tasks, then decided to go with the Workflow abstraction instead.

packages/angular_devkit/schematics/tools/workflow/node-workflow.ts
hansl: It's an interesting proposal. I think it could definitely work. I've thought about having the workflow as the place you could add sinks, but there's more implications than just that.

In general I'd say that's post 6.0 work. Having the Workflow abstraction would allow us to refactor this anyway.

packages/angular_devkit/schematics/src/engine/engine.ts
hansl: Post 6.0. I put this property as `@internal` so at least it won't be in the `.d.ts` files.

packages/angular_devkit/schematics/tools/workflow/node-workflow.ts
clydin: Yes. I agree. It was more of a brainstorming comment then anything else.

packages/angular_devkit/schematics/src/engine/task.ts
clydin: hmm...I'd probably go with removing it for now from this PR.  The underlying class is still there so it's just the _glue_ code here and we can add that in again moving forward.

hansl: Done.

PR #468, opened 2018-02-26
fix(@angular-devkit/core): resolving global npm modules was not working is some situations

ymeine:
Examples:

- when setting custom npm prefix
- when using nodist under windows (it is actually also setting custom npm prefix)

Therefore, use `npm config get prefix` to get the actual prefix, keep the environment variable `PREFIX` priority and fallback to other methods if not working.

### DIFF
--- a/packages/angular_devkit/core/node/resolve.ts
+++ b/packages/angular_devkit/core/node/resolve.ts
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import * as childProcess from 'child_process';
 import * as fs from 'fs';
 import * as path from 'path';
 import { BaseException } from '../src';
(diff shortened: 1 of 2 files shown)


hansl (dismissed):

hansl:
Please fix the test error (seems like there's a last case missing?) and this can be merged.

packages/angular_devkit/core/node/resolve.ts
```diff
@@ -50,17 +51,22 @@ function _getGlobalNodeModules() {
...
-      globalPrefix = path.join(destdir, globalPrefix);
+    const npmPrefix = _getNpmPrefix();
+    if (npmPrefix != null) {
+      globalPrefix = npmPrefix;
+    } else if (process.platform === 'win32') {
+      // c:\node\node.exe --> prefix=c:\node\
+      globalPrefix = path.dirname(process.execPath);
+    } else {
+      // /usr/local/bin/node --> prefix=/usr/local
+      globalPrefix = path.dirname(path.dirname(process.execPath));
```
hansl: I think this might be why it's failing on circle. Their execPath might be different.

ymeine:
Hello, 

I know I come back quite late, unfortunately I don't have time to work on this task lately. I did a specific commit to log some information and here it is: 

- `which node` ⇒ `/usr/local/bin/node`
- `node -pe "process.execPath"` ⇒ `/usr/local/bin/node`
- `which npm` ⇒ `/usr/local/bin/npm`
- `npm config get prefix` ⇒ `/home/circleci/.npm-global`
- `node -pe "try {require.resolve('npm');} catch (e) {''}"` ⇒ nothing
- `ls /usr/local/lib/node_modules` ⇒ gives a single entry `npm`
- `ls /home/circleci/.npm-global/node_modules` ⇒ triggers an access error since the folder doesn't exist

How to interpret it: 

- the `node` binary is in folder `/usr/local/bin` and so is the `npm` one
- the global modules path as configured through npm is `/home/circleci/.npm-global` but it seems to have no module installed in it (since there is no `node_modules` folder inside)
- resolving `npm` from Node.js code doesn't give anything, which is consistent with the fact there is apparently no module installed globally
- however we can see there there is actually a `npm` folder in `/usr/local/lib/node_modules` and that's why the test used to pass since that's an implemented use case (but the code I added takes precedence and makes it fail)

Leaving this comment here for now, hoping one could tackle this in the future.

alexeagle:
Sorry @ymeine  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #470, opened 2018-02-26
Add Schematic to for PWA's

Brocco:


packages/schematics/angular/pwa/files/assets/manifest.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "<%= title %>",
+  "short_name": "<%= title %>",
+  "theme_color": "#1976d2",
+  "background_color": "#fafafa",
+  "display": "browser",
```
cyrilletuzi: It should be `standalone`, not `browser`, otherwise it's not a PWA.

packages/schematics/angular/pwa/files/assets/manifest.json
```diff
@@ -0,0 +1,52 @@
...
+  "Scope": "/",
```
cyrilletuzi: `scope` in lowercase

packages/schematics/angular/service-worker/files/ngsw-config.json
```diff
@@ -0,0 +1,27 @@
...
+        "/*.chunk.js"
+      ]
+    }
+  }, {
+    "name": "assets",
+    "installMode": "lazy",
+    "updateMode": "prefetch",
+    "resources": {
+      "files": [
+        "/assets/**"
```
cyrilletuzi: As `assets` are lazy-loaded, is it the good place to put the `manifest.json`, which should be prefetched ? It's not an asset, it's a config file for the app like `ngsw-config.json`, so `src` would be a more consistent place to put it. But in that case, don't forget to copy it in `dist` on build.

packages/schematics/angular/pwa/files/assets/manifest.json
```diff
@@ -0,0 +1,52 @@
+{
```
cyrilletuzi: Also, I don't see the `manifest.json` included in `index.html` via `<link rel="manifest">`, is this normal ?

packages/angular/pwa/pwa/index.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export default function (options: PwaOptions): Rule {
+  return (host: Tree, context: SchematicContext) => {
+    const workspace = getWorkspace(host);
+    const project = workspace.projects[options.project];
+
+    const assetPath = join(project.root as Path, 'src', 'assets');
+
+    const tempalteSource = apply(url('./files/assets'), [
```
hansl: Typo.

packages/angular/pwa/collection.json
```diff
@@ -0,0 +1,11 @@
+{
+  "schematics": {
+    "ng-add": {
+      "factory": "./pwa",
+      "description": "Update an application with PWA defaults.",
+      "schema": "./pwa/schema.json",
+      "private": true,
```
hansl: Nit: Shouldn't be private.

cyrilletuzi:
My reviews weren't taken into account (`manifest.json` into `assets` so lazy-loaded instead of prefetched and `display` in `browser` mode instead of `standalone`), hope this won't cause problems on release...

PR #472, opened 2018-02-27
feat(@angular-devkit/core): add support for smart defaults in the Core Schema Registry

hansl:


packages/angular_devkit/core/src/json/schema/registry_spec.ts
```diff
@@ -251,4 +251,56 @@ describe('CoreSchemaRegistry', () => {
...
   });
+
+  it('supports smart defaults', done => {
+    const registry = new CoreSchemaRegistry();
+    const data: any = {
+      arr: [{}],
+    };
+
+    registry.addSmartDefaultProvider('test', () => true);
+    registry.addSmartDefaultProvider('test2', (x) => x['blue']);
```
filipesilva: Can you add a test for the absolute pointer? I think it's supported from the code, but not tested. 

Relative pointers are not supported, right?

filipesilva: Discussed offline.

PR #473, opened 2018-02-27
NG UPDATE schematics

hansl:


docs/specifications/update.md
```diff
@@ -0,0 +1,191 @@
+# Update Command
+
+`ng update` is a new command in the CLI to update one or multiple packages, its peer dependencies, and the peer dependencies that depends on it.
+
+If there are inconsistencies, for example if peer dependencies cannot be matches by a simple semver range, the tool will error out (and nothing will be committed on the filesystem).
+
+## Glossary
+
+* **migration**. A schematic that is performed on the
```
Brocco: Sentence fragment.
If there are no other terms for the glossary consider omitting it.

docs/specifications/update.md
```diff
@@ -0,0 +1,191 @@
...
+        showError('Cannot update dependency "%s": "%s" depends on an incompatible peer dependency' % (x, other))
+
+    update_package_json(p, x)
+```
+
+
+
+## Library Developers
+
+Libraries are responsible for defining their own update schematics. The `ng update` tool will update the package.json, and if it detects a special key in package.json of the library, will run the update schematic on it (with version information metadata).
```
Brocco: "special key" indicates some sort of secret here, either use it here or mention that it is described below.

docs/specifications/update.md
hansl: Removed the whole Glossary.

docs/specifications/update.md
hansl: Clarified.

PR #478, opened 2018-03-05
feat(@angular-devkit/core): support coercion in schema registry

filipesilva:


filipesilva:
@hansl done. We might have to start using an options object instead of just ordered arguments for the registry soon. It would need a minor version.

PR #495, opened 2018-03-09
refactor(@angular-devkit/build-webpack): update autoprefixer to v8.1

clydin:
browserslist defaults are based on the Angular support browsers: https://angular.io/guide/browser-support


hansl:
Maybe I missed the discussion surrounding this but why do we need this file?  It's highly confusing and it's not in a format people are used to (or any format really).

PR #497, opened 2018-03-09
feat(@angular-devkit/architect): expose builder schema

filipesilva:


filipesilva:
Example usage of getting schema prior to running a target:
```
    const target = architect.getTarget();
    architect.getBuilderDescription(target).pipe(
      concatMap(builderDescription => {
        console.log(builderDescription.schema);

        return architect.run(target);
      }),
    );
```

PR #500, opened 2018-03-11
fix(@angular-devkit/architect-cli): fix complete exit code

filipesilva:


filipesilva:
Included in https://github.com/angular/devkit/pull/535.

PR #501, opened 2018-03-11
feat(@schematics/angular): Simplify scheamtic path options

Brocco:


Brocco:
This PR requires upon angular/angular-cli#9939

packages/schematics/angular/utility/parse-name.ts
```diff
@@ -0,0 +1,27 @@
...
+// import { relative, Path } from "../../../angular_devkit/core/src/virtual-fs";
+import { Path, normalize } from '@angular-devkit/core';
+
+export interface Location {
+  name: string;
+  path: Path;
+}
+
+export function parseName(path: string | Path, name: string): Location {
+  const splitName = name.split('/');
```
hansl: There's a `split()` function in `path.ts`. Please use that.

packages/schematics/angular/utility/parse-name.ts
```diff
@@ -0,0 +1,27 @@
...
+  const nameWithoutPath = splitName.pop() || name;
```
hansl: `basename(name)`?

hansl:
Please fix my comments.

packages/schematics/angular/application/files/tsconfig.spec.json
```diff
@@ -1,7 +1,7 @@
 {
-  "extends": "<%= sourcedir.split('/').map(x => '..').join('/') %>/tsconfig.json",
+  "extends": "<%= newProjectRoot.split('/').map(x => '..').join('/') %>/tsconfig.json",
```
hansl: shouldn't that include `sourcedir`?

packages/schematics/angular/application/index.ts
```diff
@@ -21,89 +23,113 @@ import {
...
+    const projects: JsonObject = (<any> workspaceAst.value).projects || {};
+    const rootPath = `${newProjectRoot}/${options.name}`;
+    const parentDirMap = rootPath.split('').filter(x => x === '/').map(_ => '../');
+    const project = {
+      root: rootPath,
+      projectType: 'application',
+      cli: {},
+      schematics: {},
+      architect: {
+        targets: {
```
filipesilva: There is no `targets` here, what's inside should go up a level.

packages/schematics/angular/application/index.ts
```diff
@@ -21,89 +23,113 @@ import {
...
+          build: {
```
filipesilva: Can you update this to use all the default targets that are currently in https://github.com/angular/devkit/blob/5a8d4ee00eedccd068486bd158f64b0027ba14a1/packages/schematics/angular/application/files/__dot__angular.json#L13-L120 ?

packages/schematics/angular/workspace/files/package.json
```diff
@@ -12,24 +12,23 @@
...
-    "@angular/common": "^5.2.0",
-    "@angular/compiler": "^5.2.0",
-    "@angular/core": "^5.2.0",
-    "@angular/forms": "^5.2.0",
-    "@angular/http": "^5.2.0",
-    "@angular/platform-browser": "^5.2.0",
-    "@angular/platform-browser-dynamic": "^5.2.0",
-    "@angular/router": "^5.2.0",<% if (serviceWorker) { %>
-    "@angular/service-worker": "^5.2.0",<% } %>
+    "@angular/animations": "^6.0.0",
```
filipesilva: Using these versions will break CLI e2e, since they are not out yet.

packages/schematics/angular/application/index.ts
```diff
@@ -21,89 +23,113 @@ import {
...
+                plugin: '@angular-devkit/architect-transforms:replacement',
+                file: 'environments/environment.ts',
+                configurations: {
+                  production: 'environments/environment.prod.ts',
+                },
+              },
+            ],
+            options: {
+              outputPath: `${parentDirMap}/dist`,
+              index: 'index.html',
```
filipesilva: All file paths are root relative, so these need to be prefixed by `rootPath`. Have a look at the existing `__dot__angular.json` if you're unsure which need the root dir, it currently has `<%= sourceDir %>` where needed.

packages/schematics/angular/application/files/tsconfig.spec.json
Brocco: I added a couple of tests, it should be good.

packages/schematics/angular/class/schema.json
```diff
@@ -12,20 +12,11 @@
       "type": "string",
       "format": "path",
       "description": "The path to create the class.",
-      "default": "app",
       "visible": false
     },
-    "sourceDir": {
```
filipesilva: Removing options is a breaking change, we will need a new minor on everything. Not that we didn't before, this isn't the first change... but I just want to highlight that.

packages/schematics/angular/service/index.ts
```diff
@@ -48,16 +49,17 @@ function stripTsExtension(path: string): string {
...
-    throw new SchematicsException(`sourceDir option is required.`);
-  }
-
   return (host: Tree, context: SchematicContext) => {
     let providedByModule = '';
     let providedInPath = '';
 
+    if (options.path === undefined) {
+      // TODO: read this default value from the config file
+      options.path = 'src/app';
```
filipesilva: I take it this needs to be updated after another PR?

packages/angular_devkit/core/src/workspace/workspace-schema.ts
```diff
@@ -0,0 +1,101 @@
...
+   */
+  newProjectRoot?: string;
+  /**
+   * Tool options.
+   */
+  cli?: {
+    /**
+     * Link to schema.
+     */
+    $schema?: string;
```
filipesilva: I thought these did something but they don't. Only the toplevel `$schema` works. Sorry.

packages/schematics/angular/application/index_spec.ts
```diff
@@ -16,108 +17,92 @@ describe('Application Schematic', () => {
...
-    expect(files.indexOf('/foo/src/environments/environment.prod.ts')).toBeGreaterThanOrEqual(0);
-    expect(files.indexOf('/foo/src/environments/environment.ts')).toBeGreaterThanOrEqual(0);
-    expect(files.indexOf('/foo/src/app/app.module.ts')).toBeGreaterThanOrEqual(0);
-    expect(files.indexOf('/foo/src/app/app.component.css')).toBeGreaterThanOrEqual(0);
-    expect(files.indexOf('/foo/src/app/app.component.html')).toBeGreaterThanOrEqual(0);
-    expect(files.indexOf('/foo/src/app/app.component.spec.ts')).toBeGreaterThanOrEqual(0);
-    expect(files.indexOf('/foo/src/app/app.component.ts')).toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/projects/foo/environments/environment.prod.ts'))
+      .toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/projects/foo/environments/environment.ts')).toBeGreaterThanOrEqual(0);
```
filipesilva: Most of these files are source files and should be inside `src/`, with the `app` folder still existing. Only configuration files (`karma.conf.js`, `tsconfig.app.json` and `tsconfig.spec.json`) remain outside source.

You can see the target structure here (replace `projects/app` with `projects/foo`): https://github.com/filipesilva/workspace-project/tree/master/projects/app

packages/schematics/angular/application/index_spec.ts
```diff
@@ -16,108 +17,92 @@ describe('Application Schematic', () => {
...
+    expect(files.indexOf('/projects/foo/favicon.ico')).toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/projects/foo/index.html')).toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/projects/foo/karma.conf.js')).toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/projects/foo/main.ts')).toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/projects/foo/polyfills.ts')).toBeGreaterThanOrEqual(0);
+    expect(files.indexOf('/projects/foo/protractor.conf.js')).toBeGreaterThanOrEqual(0);
```
filipesilva: `protractor.conf.js` is a configuration file for the `e2e` project. You can find the intended structure here: https://github.com/filipesilva/workspace-project/tree/master/projects/app-e2e

I think we should just call it `e2e` instead of `app-e2e` though. It's cleaner and if users want more e2e project they can rename it.

packages/schematics/angular/application/other-files/app.module.ts
```diff
@@ -2,20 +2,15 @@ import { BrowserModule } from '@angular/platform-browser';
...
-import { environment } from '../environments/environment';<% } %>
 
 @NgModule({
   declarations: [
     AppComponent
   ],
   imports: [
     BrowserModule<% if (routing) { %>,
-    AppRoutingModule<% } %><% if (serviceWorker) { %>,
-    ServiceWorkerModule.register('/ngsw-worker.js', { enabled: environment.production })<% } %>
```
filipesilva: Is service worker registration somewhere else now?

packages/schematics/angular/ng-new/index.ts
```diff
@@ -0,0 +1,77 @@
...
+      inlineTemplate: options.inlineTemplate,
+      viewEncapsulation: options.viewEncapsulation,
+      routing: options.routing,
+      style: options.style,
+      skipTests: options.skipTests,
+    };
+
+    return chain([
+      schematic('workspace', workspaceOptions),
+      schematic('application', applicationOptions),
```
filipesilva: Adding the `e2e` project will probably need a separate schematic.

packages/angular_devkit/core/src/workspace/workspace-schema.ts
Brocco: This interface is only a helper, so no worries.

packages/schematics/angular/application/other-files/app.module.ts
Brocco: Not in this PR, but it is in the PWA PR #470

PR #505, opened 2018-03-13
refactor(@angular-devkit/build-webpack): remove deploy URL option and CSS special cases

clydin:
Architect version of https://github.com/angular/angular-cli/pull/9910

For reference, a comparison of the deploy URL and base HREF options can be found here: https://github.com/angular/angular-cli/blob/master/docs/design/deployurl-basehref.md

BREAKING CHANGE:

The deploy URL option has been removed. The base HREF option alone covers a majority of the use cases of the deploy URL option with the added benefit of affecting relative URL assets referenced in component HTML templates. For more advanced use cases, the Angular APP_BASE_HREF token can be used in conjuction to adjust the Angular routing root path.
Also, root-relative CSS resources will no longer be modified when a base HREF is present. For cases where a relative URL referenced resource should not be processed by the CLI, a caret ^ may be prefixed to the URL of the CSS resources to prevent the CLI from processing or otherwise modifying the URL and resource.


alan-agius4:
This change is a bit of a bummer to be honest. Which in my case depend on quite a bit. I am using the deployUrl to set the CDN url for my assets and bundles.

alexeagle:
@clydin do we still need this? Closing since it's on the devkit repo.

PR #510, opened 2018-03-14
feat(@schematics/angular): Add schematic for ng update migrations

Brocco:


packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -0,0 +1,267 @@
...
+    host.create(configPath, JSON.stringify(config, null, 2));
+
+    return host;
+  };
+}
+
+// tslint:disable-next-line:no-any
+function extractCliConfig(_config: any): JsonObject {
+  // tslint:disable-next-line:no-any
+  const cfg: any = {};
```
filipesilva: Do we have no CLI config to migrate over? Defaults for warnings and such.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -0,0 +1,267 @@
...
+function extractArchitectConfig(_config: any): JsonObject {
+  return {};
+}
+// tslint:disable-next-line:no-any
+function extractProjectsConfig(config: any): any {
+  const builderPackage = '@angular-devkit/build-webpack';
+  // convert the apps to projects
+  // tslint:disable-next-line:no-any
+  const projectMap = config.apps.map((app: any, idx: number) => {
+    const name = app.name || `$$proj${idx}`;
```
filipesilva: Maybe we should review this name for the first app, which is the one most people have. So

```
const name = app.name || idx === 0 ? 'app' : `$$proj${idx}`;
```

WDYT?

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -0,0 +1,267 @@
...
+
+    testOptions.assets = (app.assets || []).map((asset: string | JsonObject) =>
+      typeof asset === 'string' ? { glob: asset } : asset);
+
+    architect['test'] = {
+      builder: `${builderPackage}:karma`,
+      options: testOptions,
+    };
+
+    // // Protractor target
```
filipesilva: The protractor target still needs to added in a separate app:

https://github.com/filipesilva/workspace-project/blob/17b36374b76112908348c6bb08b434420e6cffc3/.workspace.json#L112-L133

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -0,0 +1,384 @@
...
+
+      // Extract target
+      const extractI18nOptions: JsonObject = { browserTarget: `${name}:build` };
+      architect['extract-i18n'] = {
+        builder: `${builderPackage}:extract-i18n`,
+        options: extractI18nOptions,
+      };
+
+        // TODO: decide how to handle config.test or config.test.karma are not defined
+      const karmaConfig = config.test && config.test.karma
```
filipesilva: If they are not defined, the old CLI could not have used them at all. So not using them in the workspace is the right approach.

packages/schematics/angular/migrations/update-6/index_spec.ts
```diff
@@ -0,0 +1,638 @@
...
+
+      it('should set build target', () => {
+        tree.create(oldConfigPath, JSON.stringify(baseConfig, null, 2));
+        tree = schematicRunner.runSchematic('migration-01', defaultOptions, tree);
+        const build = getConfig(tree).projects['app1'].architect.build;
+        expect(build.builder).toEqual('@angular-devkit/build-webpack:browser');
+        expect(build.options.scripts).toEqual([]);
+        expect(build.options.styles).toEqual([{ input: 'src/styles.css' }]);
+        expect(build.options.assets).toEqual([{ glob: 'src/assets' }, { glob: 'src/favicon.ico' }]);
+        const prodConfig = build.configurations.production;
```
filipesilva: Note: you can actually use Jasmine's `toEqual` to compare objects, and it will tell you if there are missing/extra members. 

I did it in https://github.com/angular/angular-cli/blob/5716621e3e5d42007c4109201c03705c49e445b5/packages/%40angular/cli/utilities/build-webpack-compat.spec.ts#L234-L242. 

Not a change request, just a comment.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -0,0 +1,384 @@
...
+        return entry;
+      };
+
+      const architect: JsonObject = {};
+      project.architect = architect;
+
+        // Browser target
+      const buildOptions: JsonObject = {
+        // Make outputPath relative to root.
+        outputPath: relative(normalize('/' + project.root), normalize('/' + outDir)),
```
hansl: Just `outDir`, it's already the right value.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -0,0 +1,384 @@
...
+function extractProjectsConfig(config: CliConfig): JsonObject {
+  const builderPackage = '@angular-devkit/build-webpack';
+
+  const apps = config.apps || [];
+  // convert the apps to projects
+  const projectMap = apps
+    .map((app: AppConfig, idx: number) => {
+      const name = app.name || `app${idx + 1}`;
+      const outDir = app.outDir || defaults.outDir;
+      const projectRoot = app.root || defaults.appRoot;
```
hansl: `root` in old projects is `src/`, while in new projects it shouldn't be because we imply `src/` already in other parts.

Basically `ng new x` would set the project root to `projects/x/` with src being `projects/x/src/`, but using the `app.root` here makes it `src/` with src being `src/src/`. So you cannot generate any component or schematics in general if you keep this.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -0,0 +1,411 @@
...
+
+    host.overwrite(pkgPath, JSON.stringify(pkg, null, 2));
+
+    return host;
+  };
+}
+
+export default function (): Rule {
+  return (host: Tree, context: SchematicContext) => {
+    const configPath = '/.angular-cli.json';
```
hansl: It can also be other files, please use the full list.

PR #518, opened 2018-03-16
feat(@angular-devkit/schematics): generate tree-shakeable services by default

alxhub:
Change the 'service' schematic for Angular to generate a tree-shakeable
service by default. This greatly simplifies the schematic since it no longer
needs to add a provider to @NgModule.

### DIFF
--- a/packages/schematics/angular/service/index.ts
+++ b/packages/schematics/angular/service/index.ts
@@ -5,15 +5,13 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { normalize, strings } from '@angular-devkit/core';
+import { Path, normalize, strings } from '@angular-devkit/core';
 import {
   Rule,
   SchematicContext,
   SchematicsException,
   Tree,
   apply,
-  branchAndMerge,
-  chain,
   filter,
   mergeWith,
   move,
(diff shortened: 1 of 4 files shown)


packages/schematics/angular/service/index.ts
```diff
@@ -22,61 +20,61 @@ import {
...
 import * as ts from 'typescript';
-import { addProviderToModule } from '../utility/ast-utils';
-import { InsertChange } from '../utility/change';
+import { getFirstNgModuleName } from '../utility/ast-utils';
 import { buildRelativePath, findModuleFromOptions } from '../utility/find-module';
 import { Schema as ServiceOptions } from './schema';
 
+function getModuleNameFromPath(host: Tree, modulePath: Path) {
+  if (!host.exists(modulePath)) {
+    throw new Error('Specified module does not exist');
```
Brocco: This should throw a `SchematicsException`

packages/schematics/angular/service/index_spec.ts
```diff
@@ -42,20 +42,21 @@ describe('Pipe Schematic', () => {
...
   });
 
-  it('should import into a specified module', () => {
+  it('should import a specified module', () => {
     const options = { ...defaultOptions, module: 'app.module.ts' };
 
     const tree = schematicRunner.runSchematic('service', options, appTree);
-    const content = getFileContent(tree, '/src/app/app.module.ts');
-    expect(content).toMatch(/import { FooService } from '.\/foo\/foo.service'/);
+    const content = getFileContent(tree, '/src/app/foo/foo.service.ts');
```
Brocco: Please replace usages of `getFileContent(tree, 'foo')` to `tree.readContent('foo')`. This means defining the type of `tree` to be `UnitTestTree` as `readContent` is a test helper method in a better place.

alxhub: Done.

packages/schematics/angular/service/index.ts
alxhub: Done.

PR #525, opened 2018-03-18
Webpack + VFS = <3

hansl:


packages/angular_devkit/build_webpack/src/browser/index.ts
```diff
@@ -150,6 +152,12 @@ export class BrowserBuilder implements Builder<BrowserBuilderOptions> {
           return;
         }
         const webpackCompiler = webpack(webpackConfig);
+
+        // TODO: fix webpack typings.
+        // tslint:disable-next-line:no-any
+        (webpackCompiler as any).inputFileSystem = new WebpackFileSystemHostAdapter(
+          new NodeJsSyncHost(),
```
filipesilva: Shouldn't this be using the `this.context.host` instead of a new one? Or at least one based on the that one.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  readdir(path: string, callback: Callback<string[]>): void {
+    return this._doHostCall(this._host.list(normalize('/' + path)), callback);
+  }
+
+  readFile(path: string, callback: Callback<Buffer>): void {
+    const o = this._host.read(normalize('/' + path)).pipe(
+      map(content => new Buffer(content)),
```
filipesilva: This should be mapped to string via `virtualFs.fileBufferToString`.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
```diff
@@ -0,0 +1,154 @@
...
+  readJson(path: string, callback: Callback<JsonObject>): void {
+    const o = this._host.read(normalize('/' + path)).pipe(
+      map(content => JSON.parse(virtualFs.fileBufferToString(content))),
+    );
+
+    return this._doHostCall(o, callback);
+  }
+
+  readlink(path: string, callback: Callback<string>): void {
+    const err: NodeJS.ErrnoException = new Error('Not a symlink.');
```
filipesilva: We should support symlinks in the future.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
```diff
@@ -0,0 +1,154 @@
...
+      );
+
+      this._doHostCall(o, callback);
+    } else {
+      this._doHostCall(result, callback);
+    }
+  }
+
+  readdir(path: string, callback: Callback<string[]>): void {
+    return this._doHostCall(this._host.list(normalize('/' + path)), callback);
```
filipesilva: I think this will need to return the system paths instead of the normalized `Paths`.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
```diff
@@ -0,0 +1,154 @@
...
+  readJsonSync(path: string): string {
+    return JSON.parse(this.readFileSync(path));
+  }
+  readlinkSync(path: string): string {
+    const err: NodeJS.ErrnoException = new Error('Not a symlink.');
+    err.code = 'EINVAL';
+    throw err;
+  }
+
+  purge(_changes?: string[] | string): void {}
```
filipesilva: This will need to be implemented in the future.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
hansl: I disagree, at least as far as the host is concerned, symlinks were intentionally left out of the design.

We "support" them by having the Node*Host supports proper stats and you check them, but if you don't know which host comes in then you cannot know (and shouldn't IMO).

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
hansl: Maybe as input, but `readdir` returns fragments which do not contain slashes.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
hansl: I'm not sure what this should do for our own adapter. For a cache system you can clear the cache, but here I don't thikn there's anything to do.  `purge` is optional on the real typings btw.

packages/angular_devkit/build_webpack/src/browser/index.ts
hansl: Yes it should. Sorry I'll fix it. I tried it and it was failing, so decided to focus on one error at a time. Since all paths are absolute not to the Host but to the file system, that's a deeper change in general.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
hansl: Actually the webpack adapter wants to have node Buffers as a return, because it expects this to match the `fs` interface.

filipesilva: Oh, I didn't know that. In the ngtools/webpack FS we return the string. Maybe that's broken there.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
filipesilva: Let's leave it off and see if there's any problem then.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
filipesilva: Good point.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
filipesilva: Ok, I don't have any usecase that needs them right now either.

packages/angular_devkit/build_webpack/src/utils/webpack-file-system-host-adapter.ts
hansl: It's not broken for source files. The only thing Webpack is doing is `toString('utf-8')`, which works for strings :shrug:

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
```diff
@@ -193,29 +231,105 @@ export class SimpleMemoryHost implements Host<{}> {
...
+    return new Observable<boolean>(obs => {
+      obs.next(this._isFile(path));
+      obs.complete();
+    });
   }
 
-  stat(_path: Path): Observable<Stats<{}>> | null {
-    return null;
+  // Some hosts may not support stat.
+  stat(path: Path): Observable<Stats<{}>> {
```
clydin: `Observable<Stats>`?

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
```diff
@@ -193,29 +231,105 @@ export class SimpleMemoryHost implements Host<{}> {
...
+  rename(from: Path, to: Path): Observable<void> {
+    return new Observable<void>(obs => {
+      this._rename(from, to);
+      obs.next();
+      obs.complete();
+    });
+  }
 
-    return observableOf(this._cache.has(path) || this._isDir(path));
+  list(path: Path): Observable<PathFragment[]> {
```
clydin: Thoughts on this returning `Observable<PathFragment>` instead of the array.  Would allow the consumer to operator on the stream of elements directly.

packages/angular_devkit/core/node/host.ts
```diff
@@ -224,24 +223,30 @@ export class NodeJsSyncHost implements virtualFs.Host<fs.Stats> {
...
   write(path: Path, content: virtualFs.FileBuffer): Observable<void> {
-    // Create folders if necessary.
-    const _createDir = (path: Path) => {
-      if (fs.existsSync(getSystemPath(path))) {
-        return;
-      }
+    return new Observable(obs => {
+      // Create folders if necessary.
+      const _createDir = (path: Path) => {
+        if (fs.existsSync(getSystemPath(path))) {
```
clydin: Probably a refactor for the future but for all the sync calls could we use the async versions with `utils.promisify` and convert and return the _observabled_ promise?

packages/angular_devkit/build_webpack/src/browser/index.ts
```diff
@@ -150,6 +152,12 @@ export class BrowserBuilder implements Builder<BrowserBuilderOptions> {
           return;
         }
         const webpackCompiler = webpack(webpackConfig);
+
+        // TODO: fix webpack typings.
+        // tslint:disable-next-line:no-any
+        (webpackCompiler as any).inputFileSystem = new WebpackFileSystemHostAdapter(
```
clydin: The watch file system will be unaffected and I don't think caching will be used.  This webpack plugin is always applied before anything else: https://github.com/webpack/webpack/blob/master/lib/node/NodeEnvironmentPlugin.js
https://github.com/webpack/webpack/blob/master/lib/webpack.js#L34

packages/ngtools/webpack/src/gather_diagnostics.ts
```diff
@@ -56,29 +56,29 @@ export function gatherDiagnostics(
     const tsProgram = program as ts.Program;
     // Check syntactic diagnostics.
     time(`${benchmarkLabel}.gatherDiagnostics.ts.getSyntacticDiagnostics`);
-    checkDiagnostics(tsProgram.getSyntacticDiagnostics);
+    checkDiagnostics(tsProgram.getSyntacticDiagnostics.bind(tsProgram));
```
clydin: `() => tsProgram.getSyntacticDiagnostics()` ?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -87,6 +88,8 @@ export interface AngularCompilerPluginOptions {
 
   // Use tsconfig to include path globs.
   compilerOptions?: ts.CompilerOptions;
+
+  host: virtualFs.Host<fs.Stats>;
```
clydin: optional?

packages/ngtools/webpack/src/gather_diagnostics.ts
filipesilva: Yeah that looks better imho.

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
hansl: I see `Observable<X>` here as a replacement of `Promise<>`. The contract is that these will always emit once and complete, or error out. I like the consistence of this interface too.

packages/angular_devkit/core/node/host.ts
hansl: You mean the async ones? The sync ones need to be sync.

clydin: Yes. Sorry for the confusion.

packages/angular_devkit/build_webpack/test/utils/run-target-spec.ts
```diff
@@ -31,7 +32,7 @@ export function runTargetSpec(
   host: TestProjectHost,
   targetSpec: TargetSpecifier,
   overrides = {},
-  logger: logging.Logger = new logging.NullLogger(),
+  logger: logging.Logger = createConsoleLogger(),
```
filipesilva: This change should be reverted.

filipesilva:
I will followup this PR with fixes for the Windows CI.

PR #530, opened 2018-03-19
fix(@angular-devkit/core): sometimes NullLogger.createChild can be ca…

filipesilva:
…lled


filipesilva:
@googlebot I can confirm that both me and @hansl are ok with our commits being contributed to this project.

PR #531, opened 2018-03-19
feat(@angular-devkit/build-optimizer): support webpack side effects flag

clydin:


clydin:
Updated to now work on older versions of webpack.

PR #535, opened 2018-03-20
Lib integration fixes, and multiple other fixes

filipesilva:


packages/angular_devkit/build_webpack/src/browser/schema.json
```diff
@@ -29,7 +29,6 @@
     },
     "tsConfig": {
       "type": "string",
-      "default": "tsconfig.app.json",
```
hansl: Considering this is required, shouldn't that still be there?

filipesilva: They should not have defaults precisely because they are required. Their absence breaks the build.

If they had a default, they should not be required. The build would also fail while running because a file does not exist, instead of failing validation.

I think failing validation is better than failing at runtime with a message about a file that is not mentioned anywhere in the project proper.

hansl: I see. The default is wrong. Got it.

packages/schematics/angular/library/index.ts
```diff
@@ -61,42 +53,35 @@ function updateJsonFile<T>(host: Tree, path: string, callback: UpdateJsonFn<T>):
...
 
     return updateJsonFile(host, 'package.json', (json: PackageJsonPartialType) => {
+      if (!json['dependencies']) {
+        json['dependencies'] = {};
+      }
+
+      json.dependencies = {
+        '@angular/common': '^5.2.0',
+        '@angular/core': '^5.2.0',
+        '@angular/compiler': '^5.0.0',
```
clydin: These versions should be consistent and ideally not repeated to simplify updates moving forward.

packages/schematics/angular/library/index.ts
```diff
@@ -61,42 +53,35 @@ function updateJsonFile<T>(host: Tree, path: string, callback: UpdateJsonFn<T>):
...
+
       if (!json['devDependencies']) {
         json['devDependencies'] = {};
       }
 
       json.devDependencies = {
-        '@angular/compiler': '^5.0.0',
         '@angular/compiler-cli': '^5.0.0',
         'ng-packagr': '^2.2.0',
         'tsickle': '>=0.25.5',
```
clydin: `@angular/compiler-cli` has a direct dependency on `tsickle` is there a reason it is here as well?

packages/schematics/angular/library/index.ts
```diff
@@ -61,42 +53,35 @@ function updateJsonFile<T>(host: Tree, path: string, callback: UpdateJsonFn<T>):
...
         'tslib': '^1.7.1',
         'typescript': '>=2.4.2',
```
clydin: New projects should ideally use the latest supported version.

packages/schematics/angular/library/index.ts
```diff
@@ -133,7 +117,29 @@ export default function (options: GenerateLibraryOptions): Rule {
         mergeWith(templateSource),
       ])),
       options.skipPackageJson ? noop() : addDependenciesAndScriptsToPackageJson(),
-      options.skipTsConfig ? noop() : updateTsConfig(name, forwardSlashes(entryFilePath)),
+      schematic('module', {
+        name: name,
+        commonModule: false,
+        flat: true,
+        path: `/${name}/src`,
+        sourceDir: options.baseDir,
```
clydin: `sourceDir` was removed.  The path is relative to the workspace root.

packages/schematics/angular/library/index.ts
filipesilva: `ng-packagr` itself has a peer dependency on `tsickle`: https://github.com/dherges/ng-packagr/blob/master/package.json#L59-L64

packages/schematics/angular/library/index.ts
filipesilva: Will fix, this and the TS one below.

packages/schematics/angular/library/index.ts
filipesilva: I fixed that one in another commit in this PR AFAIK.

PR #541, opened 2018-03-21
fix(@angular-devkit/build-webpack): remove RxJS path mapping

sarunint:
`rxjs@6.0.0-beta.0` added back the `module` and `es2015` keys in the `package.json` and the `path-mapping.js` is retained for backward compatibility only.


filipesilva:
Heya @sarunint, I think this change makes sense but ngtools/webpack still needs to support RxJs 5.6. So I don't think we can remove it right now.

@hansl can you confirm?

clydin:
Angular 5 has a minimum rxjs version of 5.5.  Until Angular 5 support is dropped this cannot be removed.

PR #547, opened 2018-03-21
fix(@schematics/angular): Fix schematics build and generate bugs

Brocco:


packages/schematics/angular/workspace/files/package.json
```diff
@@ -41,6 +41,7 @@
     "karma-jasmine": "~1.1.1",
     "karma-jasmine-html-reporter": "^0.2.2",
     "protractor": "~5.3.0",
+    "rxjs": "6.0.0-beta.1",
```
filipesilva: Should be either one or the other. Here you have rxjs listed twice.

PR #549, opened 2018-03-22
fix(@angular-devkit/schematics): add magic tap

filipesilva:
This tap should not do anything, but if it is not here then the mergeMap below does not seem to emit anything when there are multiple copies of RxJs in node_modules.

This can be reproduced in https://github.com/filipesilva/schematics-rxjs-compat.

Running `npm run lib` will work, but if `rxjs@6.0.0-beta.1` is installed, which forces each schematics package to have unhoist its RxJs version, then schematics stops working.

This persists even after installing `rxjs@5.5.7` back, as long as `npm ls rxjs` shows multiple RxJs packages.


filipesilva:
Fixed by https://github.com/ReactiveX/rxjs/pull/3476

PR #558, opened 2018-03-23
feat(build webpack): re-enable `HtmlWebpackPlugin` as it now supports…

alan-agius4:
re-enable `HtmlWebpackPlugin` as it now support webpack 4


packages/angular_devkit/build_webpack/src/angular-cli-files/models/webpack-configs/browser.ts
```diff
@@ -21,34 +20,29 @@ import { WebpackConfigOptions } from '../build-options';
...
-        collapseWhitespace: true,
-        keepClosingSlash: true
-      } : false
-    }));
-    extraPlugins.push(new BaseHrefWebpackPlugin({
-      baseHref: buildOptions.baseHref as string
-    }));
-  }
+  extraPlugins.push(new HtmlWebpackPlugin({
+    template: path.resolve(root, appConfig.index),
```
erikyuzwa: was just looking to submit my own similar patch. Would it be possible to update this to allow for a given `appConfig.template` entry? 
```
template: path.resolve(root, appConfig.template || appConfig.index), 
```
Allows for an index template override, with a fallback to the current index entry...

alan-agius4: In all fairness I am still looking into making the CI green as I still have some tests failing.

Not quite understanding why you need two templates. If you want to override the index file. You can specify this in the cli config. by changing the `index.html` to something else.

PR #593, opened 2018-03-30
Ng update assets

hansl:


packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -176,11 +176,77 @@ function extractProjectsConfig(config: CliConfig): JsonObject {
       const outDir = app.outDir || defaults.outDir;
       const appRoot = app.root || defaults.appRoot;
 
+      function _mapAssets(asset: string | JsonObject) {
+        if (typeof asset === 'string') {
+          return { glob: asset, input: appRoot + '/', output: outDir + '/' };
```
filipesilva: Output should start with '/',

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -268,10 +315,7 @@ function extractProjectsConfig(config: CliConfig): JsonObject {
         }
       testOptions.scripts = (app.scripts || []).map(extraEntryMapper);
```
filipesilva: Scripts and styles mapping must use `bundleName` instead of `output`. For lazy ones, the bundle name must be determined manually by removing the file extension of the input.

hansl: Turns out that we didn't have an `output` for scripts and styles.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -176,25 +176,94 @@ function extractProjectsConfig(config: CliConfig): JsonObject {
...
+
+      function _extraEntryMapper(extraEntry: string | JsonObject) {
         let entry: JsonObject;
         if (typeof extraEntry === 'string') {
           entry = { input: extraEntry };
         } else {
-          entry = extraEntry;
+          entry = {
+            input: extraEntry.input,
+            bundleName: extraEntry.output,
```
filipesilva: You have to specialcase `bundleName` for extra entries that have `lazy:true`. They didn't need to specify the bundle name before because we calculated it via `basename` plus stripping the extension. But now we're doing it verbose so it needs the bundle name.

filipesilva: You also need to pass on the `lazy` property, that isn't being passed on right now.

hansl: Replaced with

```typescript
          const input = extraEntry.input as string;
          const bundleName =
            extraEntry.output
            || basename(
              normalize(input.replace(/\.(js|css|scss|sass|less|styl)$/i, '')),
            );
          entry = { input, bundleName };
```

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -176,23 +176,96 @@ function extractProjectsConfig(config: CliConfig): JsonObject {
...
         if (typeof extraEntry === 'string') {
-          entry = { input: extraEntry };
+          entry = { input: join(app.root as Path, extraEntry) };
         } else {
-          entry = extraEntry;
+          const input = join(app.root as Path, extraEntry.input as string || '');
+          const lazy = !!extraEntry.lazy;
+          entry = { input, lazy };
+
+          if (!extraEntry.output && !lazy) {
```
filipesilva: `!lazy` -> `lazy`. It only needs to be inferred for lazy entries.

hansl: Done.

PR #603, opened 2018-04-02
Fix deprecation warning with Purify Webpack plugin

mtraynham:
Update Purify plugin to fix deprecation warning with Webpack 4


eseliger:
Any chance on getting this soon? :) It's a pain to use angular with webpack 4 at the moment and this is one more step towards a stable webpack 4 build!

alexeagle:
Sorry @mtraynham  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #611, opened 2018-04-03
fix(@schematics/update): migration only should use installed as target

hansl:


packages/schematics/update/update/index.ts
```diff
@@ -248,7 +248,8 @@ function _migrate(
   if (!info) {
     return of<void>();
   }
-  const target = info.target;
+console.log(info)
```
filipesilva: console log

hansl: fixe.d

PR #625, opened 2018-04-05
feat(@schematics/angular): add zone modules and test flag

JiaLiPassion:
- upgrade `zone.js` to `0.8.26`
- add zone modules import into `polyfill.ts`, user can choose which to import based on requirements.
- add new feature flag for `fakeAsync` test.


alexeagle:
Sorry @JiaLiPassion  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #626, opened 2018-04-05
fix(@angular/pwa): typo and display standalone in manifest.json

cyrilletuzi:
In `manifest.json`: 
- fixes a typo in `scope` property
- changes `display` to `standalone` instead of `browser` (otherwise it's just a normal website, not a PWA)

I'm also concerned about the manifest being in `assets`, as the default `ngsw-config.json` tells to lazy load `assets`, while the `manifest.json` should certainly be prefetched (but as it should be load from start by the browser, maybe it's OK).

@filipesilva

### DIFF
--- a/packages/angular/pwa/pwa/files/assets/manifest.json
+++ b/packages/angular/pwa/pwa/files/assets/manifest.json
@@ -3,8 +3,8 @@
   "short_name": "<%= title %>",
   "theme_color": "#1976d2",
   "background_color": "#fafafa",
-  "display": "browser",
-  "Scope": "/",
+  "display": "standalone",
+  "scope": "/",
   "start_url": "/",
   "icons": [
     {


cyrilletuzi:
Don't think the AppVeyor issue is related to this PR.

filipesilva:
Can confirm, the appveyor failure is unrelated. It looks like a problem in our protractor tests. I'll have a look in a separate PR.

alxhub:
It would be "most correct" to cache the `manifest.json` in prefetch mode, but caching it lazily is okay - it will be loaded by the browser, so it will be cached eventually. And I believe with PWA install to home screen, the browser saves the data in the manifest independently (it needs to e.g. remember the splash screen icon and home URLs)

Brocco:
For now, I would prefer to leave the default `display` value as `browser`. It is the default value and will be the assumed default for most users.

cyrilletuzi:
Could you explain why you assume it will be the default for most users? The point of this package is to have a PWA, but with `display: browser` you just have a classic website with a shortcut, so I don't see the point.

PR #627, opened 2018-04-05
fix(@schematics/angular): remove broken imports

hansl:
Fixes https://github.com/angular/angular-cli/issues/10190

### DIFF
--- a/packages/schematics/angular/universal/index.ts
+++ b/packages/schematics/angular/universal/index.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/universal/index.ts
```diff
@@ -171,7 +174,7 @@ function addDependencies(): Rule {
   };
 }
 
-function getTsConfigOutDir(host: Tree, architect: Architect): string {
+function getTsConfigOutDir(host: Tree, architect: experimental.workspace.Architect): string {
```
filipesilva: It's odd to have something called `Architect` and `Project` being exported out of the workspace... we should rename these to `ArchitectSchema` and `ProjectSchema`.

PR #631, opened 2018-04-05
feat(@angular-devkit/schematics): add lifecycle hooks to workflow

hansl:
This will allow us to show logs at each new phases, for example, instead of only at the end.

Reference: https://github.com/angular/angular-cli/issues/10155


packages/angular_devkit/schematics/tools/workflow/node-workflow.ts
```diff
@@ -17,6 +17,7 @@ import {
   workflow,
 } from '@angular-devkit/schematics';  // tslint:disable-line:no-implicit-dependencies
 import { EMPTY, Observable, Subject, concat, of, throwError } from 'rxjs';
+import { reduce, tap } from 'rxjs/internal/operators';
```
trotyl: Any reason for importing internal?

clydin: no.  that was accidental.

PR #633, opened 2018-04-05
ci: update appveyor chrome and protractor

filipesilva:


filipesilva:
@hansl PTAL, now it is green.

PR #641, opened 2018-04-06
fix(@schematics/angular): Correct paths for linting libraries

chrste90:
When creating a new library, the paths for lint were false.

Now the correct paths should be set.

CC @filipesilva


filipesilva:
Oh, good catch! Thanks for fixing this!

PR #642, opened 2018-04-06
fix(@angular-devkit/schematics): Fix deep copy logic

Brocco:


packages/angular_devkit/core/src/utils/object.ts
```diff
@@ -14,3 +14,23 @@ export function mapObject<T, V>(obj: {[k: string]: T},
     return acc;
   }, {});
 }
+
+// tslint:disable-next-line:no-any
+export function deepCopy<T extends {[key: string]: any}> (object: T): T | T[] {
```
hansl: Remove the `extends ...` part. This is valid with any value anyway.

packages/angular_devkit/schematics/tools/schema-option-transform.ts
```diff
@@ -26,29 +27,11 @@ export class InvalidInputOptions extends BaseException {
...
-  // return copy;
-}
-
-
 // This can only be used in NodeJS.
 export function validateOptionsWithSchema(registry: schema.SchemaRegistry) {
   return <T extends {}>(schematic: SchematicDesc, options: T): Observable<T> => {
     // Prevent a schematic from changing the options object by making a copy of it.
-    options = _deepCopy(options);
+    options = deepCopy(options) as T;
```
hansl: You shouldn't need `as T` here.

Issue #570, opened 2018-03-27
CLI schematics regression: missing rules in tslint.json


cexbrayat:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Area
```
- [ ] devkit
- [x] schematics
```

### Repro steps

Just generate a project with CLI `6.0.0-beta.7`.
The `tslint.json` file now lacks the `directive-selector` and `component-selector` rules.

### Desired functionality

It should still have the rules.

### Mention any other details that might be useful

See the difference between beta.6 and beta.7 https://github.com/cexbrayat/angular-cli-diff/compare/6.0.0-beta.6...6.0.0-beta.7

cexbrayat:
I just realized that the `prefix` option is completely ignored too when generating a new project. Is it getting removed or is it not yet implemented?

filipesilva:
Regarding the selector rules being gone, that is a bug of sorts. I removed them because now you can generate apps and libraries, and linting would fail right after doing so.

The right behaviour is for app/lib generation to add those prefixes to tslint.

chrste90:
I would like to look into it but have some questions first:

1. There are currently no prefixes in angular.json for apps or libs. Should these be added back as well?
2. And how should it work if there are different prefixes for each app/lib? Then the rule would be added multiple times with different values. Does this work? Or is it possible to add more than one prefix to the rules?

chrste90:
I like the idea of different tslint configs for each app/lib.

Should the `ng lint` command find the appropriate config at runtime or should this be added to `angular.json`?

filipesilva:
The lint builder already picks the closes `tslint.json` file actually. So as long as there is one extending the base one, it should work.

chrste90:
Okay. So the only thing needed is to update the schematics to create a new tslint file which extends the base tslint and adds the two missing rules? (Aside from the missing prefix in angular.json)

sfabriece:
not sure if this  should go here: 

currently new projects have an rxjs blacklist in tslint.

the rxjs blacklist should be removed for v6, sincethe recommended way to import creators is directly from `'rxjs'`

chrste90:
It ist already fixed in master and should be in the next release: https://github.com/angular/devkit/pull/600

sfabriece:
thanks @chrste90

PR #644, opened 2018-04-06
feat(@schematics/angular): Add directive and component selector rules

chrste90:
Here is the initial implementation and fix of https://github.com/angular/devkit/issues/570

Some questions:
1. When i add the `schematics/angular/application/files/root/tslint.json`, the workspace tslint file is not copied when i run `ng new App`, only the extending tslint exists in the created application. Is it not possible to create two files with the same name in different directories or am i missing something?
1. The LibraryOptions don't have a property for prefix, so i just chose `lib` for now. Should i add a prefix property and use it?
1. Should i add tests somewhere?

CC @filipesilva

### DIFF
--- a/packages/schematics/angular/application/files/lint/__tsLintRoot__/tslint.json
+++ b/packages/schematics/angular/application/files/lint/__tsLintRoot__/tslint.json
@@ -1,0 +1,17 @@
+{
+    "extends": "<%= relativeTsLintPath %>/tslint.json",
+    "rules": {
+        "directive-selector": [
+            true,
+            "attribute",
+            "<%= prefix %>",
+            "camelCase"
+        ],
+        "component-selector": [
+            true,
+            "element",
+            "<%= prefix %>",
+            "kebab-case"
+        ]
+    }
+}
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -270,7 +270,8 @@ export default function (options: ApplicationOptions): Rule {
       throw new SchematicsException(`Invalid options, "name" is required.`);
     }
     validateProjectName(options.name);
-    const appRootSelector = `${options.prefix || 'app'}-root`;
+    const prefix = options.prefix || 'app';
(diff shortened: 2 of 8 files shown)


filipesilva:
@Brocco could you have a look please? Mike really is the authority on these things so I'd defer to him.

chrste90:
@Brocco I made all the changes.

Only the problem 1 persists and i don't really know what i can do.

As soon as i add `schematics/angular/application/files/root/tslint.json`, the `schematics/angular/workspace/files/tslint.json` isn't copied any more to a new angular project. If i rename the `schematics/angular/application/files/root/tslint.json` to `tslint2.json`, the workspace file is copied again. Any idea what the problem could be?

filipesilva:
@chrste90 now that you mention it again, I think I ran into a similar problem in the library schematics... 
https://github.com/angular/devkit/blob/master/packages/schematics/angular/library/index.ts#L181-L190

Having moved files with the same name should work, but is bugged and does not. The workaround is to have a template folder named `__something__` with the problematic file there, passing the `something` var to the schematics. Schematics will replace that folder name with the path in the variable. 

When I did that I removed the `move` in favor of using the template folder but I think it is more correct to leave it in, and only add the template folder for the problematic file.

chrste90:
Thanks @filipesilva 
I made the changes and it seems that everything is working now.
So i think the changes are ready to review.

chrste90:
Also another thing (maybe you know already):
When i run the tests locally on my macbook, one test fails:
```
1) NodeJsSyncHost can watch
  Message:
    Expected 0 to be 3.
  Stack:
    Error: Expected 0 to be 3.
        at Promise.resolve.then.then.then.then.then.then.then.then (/Users/chris/Desktop/devkit/packages/angular_devkit/core/node/file:/Users/chris/Desktop/devkit/packages/angular_devkit/core/node/host_spec.ts:101:34)
        at <anonymous>
```

filipesilva:
@chrste90 yeah that test is flaky on some environments. Not too sure why myself, but it is known.

PR #646, opened 2018-04-06
fix(@schematics/angular): Allow for scoped library names

Brocco:
fixes angular/angular-cli#10172

### DIFF
--- a/packages/schematics/angular/library/index.ts
+++ b/packages/schematics/angular/library/index.ts
@@ -28,6 +28,7 @@ import {
   getWorkspace,
 } from '../utility/config';
 import { latestVersions } from '../utility/latest-versions';
+import { validateProjectName } from '../utility/validation';
 import { Schema as LibraryOptions } from './schema';
 
 
@@ -125,7 +126,7 @@ function addDependenciesToPackageJson() {
 }
 
 function addAppToWorkspaceFile(options: LibraryOptions, workspace: WorkspaceSchema,
-                               projectRoot: string): Rule {
+                               projectRoot: string, packageName: string): Rule {
 
   const project: WorkspaceProject = {
     root: `${projectRoot}`,
@@ -166,7 +167,7 @@ function addAppToWorkspaceFile(options: LibraryOptions, workspace: WorkspaceSche
     },
   };
 
-  return addProjectToWorkspace(workspace, options.name, project);
+  return addProjectToWorkspace(workspace, packageName, project);
 }
 
 export default function (options: LibraryOptions): Rule {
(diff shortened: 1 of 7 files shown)


packages/schematics/angular/library/index.ts
```diff
@@ -171,7 +171,11 @@ export default function (options: LibraryOptions): Rule {
     if (!options.name) {
       throw new SchematicsException(`Invalid options, "name" is required.`);
     }
-    const name = options.name;
+    // If scoped project (i.e. "@foo/bar"), convert projectDir to "foo-bar".
+    const packageName = options.name;
+    options.name = /@[\S]*\/[\S]*/.test(options.name)
```
hansl: What if the name passed on the command line is `a/b/c`? Is it going to create `c` in `/projects/a/b`?

hansl:
The removal of `@` and replacing `/` with `-` looks good, I just want to make sure we still allow for other cases.

filipesilva:
I think it would be good to keep the `/`. So `@scoped/lib` would go into `projects/scoped/lib`, kinda like we have things in the CLI and DevKit repos. No opinion on the `@` myself.

Brocco:
@about-code in my experience having `@` in paths messes up autocompleting paths in consoles (both Mac & Windows) which is why I removed it as part of the directory structure.

Brocco:
@hansl I updated the logic to validate the project name... but not via a schema pattern in order to provide more verbose error messages.

about-code:
I am not yet sure whether it's preferable to strip the `@` even though I see the advantages on the console. I certainly don't know enough about the full library implementation in the CLI to make any concrete suggestions. But let me explain some thoughts I had, e.g. about how it might affect webpack module resolution and potential benefits for lazy loaded modules: 

Assume an Angular app wants to lazy load an NgModule from a library in the `projects` folder. If Webpack were configured to [resolve import paths](https://webpack.js.org/concepts/module-resolution/#module-paths) from the *projects* folder like so:

```
module: { 
     resolve: ["./projects", "./node_modules"]
}
```
then it *might* (not tested in CLI context) become possible to compile angular apps which load an NgModule from a library with an absolute path beginning with the package name maybe like so:

```
// Assumption: AoT-Compiler translates the path to a dynamic webpack import()
loadChildren: '@foo/bar/src/lib/bar.module#BarModule' 
```

Webpack were able to resolve `import("@foo/bar")` by not just looking into `node_modules` but also in `projects` which may avoid the need for npm linking, or copying things into `node_modules` in some situations.  I am not sure how CLI projects are configured in this regard but I think it could have some benefits to take npm module resolution algorithms into account.

about-code:
@filipesilva Thank you very much for your response and for taking away my concerns. Now I am gonna be quiet on this. :smiley:

packages/schematics/angular/utility/validation.ts
```diff
@@ -25,3 +25,52 @@ export function validateHtmlSelector(selector: string): void {
...
+    const msg = tags.stripIndent`
+    ${firstMessage}
+    ${projectName}
+    ${Array(errorIndex + 1).join(' ') + '^'}
+    `;
+    throw new SchematicsException(msg);
+  } else if (unsupportedProjectNames.indexOf(projectName) !== -1) {
+    throw new SchematicsException(`Project name "${projectName}" is not a supported name.`);
+  } else if (!packageNameRegex.test(projectName)) {
+    throw new SchematicsException(`Project name (${projectName}) is invalid.`);
```
hansl: `${JSON.stringify(projectName)}`

rodchenkov:
Hi All
Just found that in tsconfig.json path name for scoped library is without scope name.
I think we need to solve this.

filipesilva:
@rodchenkov I have a fix incoming in https://github.com/angular/devkit/pull/895

PR #654, opened 2018-04-09
feat(@schematics/angular): simplify service test

cexbrayat:
This is a proposal to simplify the schematic for the service test.

Now that the service schematic uses `providedIn` by default, we can simplify the test to use `TestBed.configureTestingModule({})`.

Also we can use `TestBed.get()` instead of the `inject` wrapper.


cexbrayat:
I updated the PR now that https://github.com/angular/devkit/pull/687 from @Brocco has landed, and removed the `providedIn` option for services. The service test should now reflect that.

cexbrayat:
@Brocco @hansl Is there something I need to do to land this PR?

hansl:
@cexbrayat No, but this should be in the next major, not in a patch or minor.

alexeagle:
Sorry @cexbrayat  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #655, opened 2018-04-09
refactor(@schematics/angular): bump TS package dependency version to 2.7

### LINKED ISSUE #651: @schematics/angular typescript dependency is still 2.6
### Bug Report or Feature Request
```
- [x] bug report
- [ ] feature request
```

### Area
```
- [ ] devkit
- [x] schematics
```

### Versions
`node --version` : v9.11.1
`npm --version` : v5.6.0
Linux Debian 9.4

### Repro steps
```
git clone git@github.com:noelmace/nx.git
cd nx
git checkout 9c1b1bf0aa2410c6ba32
yarn install
yarn build
```

### The log given by the failure
```
packages/bazel/src/utils/ast-utils.ts(32,38): error TS2345: Argument of type 'ts.SourceFile' is not assignable to parameter of type 'ts.SourceFile'.
  Types of property 'kind' are incompatible.
    Type 'ts.SyntaxKind.SourceFile' is not assignable to type 'ts.SyntaxKind.SourceFile'. Two different types with this name exist, but they are unrelated.
packages/bazel/src/utils/ast-utils.ts(206,38): error TS2345: Argument of type 'ts.SourceFile' is not assignable to parameter of type 'ts.SourceFile'.
packages/bazel/src/utils/ast-utils.ts(233,32): error TS2345: Argument of type 'ts.SourceFile' is not assignable to parameter of type 'ts.SourceFile'.
```

### Desired functionality
Upgrade @schematics/angular typescript dependency to 2.7.2 in order to play well with angular v6 and therefore @angular/cli v6.

See https://github.com/noelmace/nx/commit/9c1b1bf0aa2410c6ba32c7787006c847bb8faf40 for the complete list of conflicting dependencies.

The code source corresponding to the failure's log is https://github.com/noelmace/nx/blob/9c1b1bf0aa2410c6ba32c7787006c847bb8faf40/packages/bazel/src/utils/ast-utils.ts

``yarn list --pattern typescript`` : 
```
yarn list v1.5.1
├─ @schematics/angular@0.5.4
│  └─ typescript@2.6.2
└─ typescript@2.7.2
Done in 1.11s.
```

In the mean time, I'm pretty sure their was a workaround for this kind of typings conflict, but I can't remember one. If someone knows and could share one, it would frees my mind (even if it's not necessary given that the upgrade is needed anyway).

#### investigations

So, I tried to look at this issue by myself, and found that : 
1. TypeScript ~2.6.2 was added to dependencies in v0.4.2 (by https://github.com/angular/devkit/commit/727e179052c264dd60357dba26e1bf362b5a76a6#diff-f1300208a045563d7d86dd174a32c14d) in order to resolve https://github.com/angular/devkit/issues/365
2. Angular-cli still rely on TS 2.6 in its devDependencies, but installing a local @schematics/angular with a TS dependency version set to 2.7 doesn't seems to breack any test
3. bumping TS dependency version in schematics/angular package.json doesn't  make any failure in devkit tests

So I'll make a minimal PR for that, and wait for feedbacks :slightly_smiling_face:

noelmace:
should close #651 

Tested with yarn test and in angular/cli tests.

### DIFF
--- a/packages/schematics/angular/package.json
+++ b/packages/schematics/angular/package.json
@@ -14,6 +14,6 @@
   "dependencies": {
     "@angular-devkit/core": "0.0.0",
     "@angular-devkit/schematics": "0.0.0",
-    "typescript": "~2.6.2"
+    "typescript": ">=2.6.2 <2.8"
   }
 }


noelmace:
circleci : "The following commit should not have a scope" ... I don't understand why. Should I remove the scope because this commit doesn't change any source file in this package ?

clydin:
The package is already being built with TS 2.7 so `build` is not appropriate in this case as TS is being used here as a library not a build tool.  It should also build fine with bazel as well since the compiling version of TS should not be used to determine type information for the actual dependency version.

But to answer your question, yes, build should not have a scope as it is meant to reflect the build process for the entire repository.

noelmace:
Ok so, about the commit conventions : should I use feat or refactor :sweat_smile: ?

As I'm not familiar with devkit and schematics, I tought it was also used as a build tool somehow, my bad.

Given that it's only the library part which is impacted by this PR, I suppose this should be ok.

Also, as ``@schematics/angular`` is one of the few devkit packages that doesn't import typescript, maybe I could simply remove the dependency entry here, instead of changing the version.

Angular-cli doesn't seems to use it neither. Only nrwl/nx ... So, why should we keep Typescript as a dependency here ?

clydin:
`refactor` would probably be the more appropriate option as no new features are really being added.

As to TypeScript's usage, the AST API is used within the Angular schematics to find and update existing TS files (e.g., adding a component to its parent module).  For reference, here is some of the utility code: https://github.com/angular/devkit/blob/master/packages/schematics/angular/utility/ng-ast-utils.ts

PR #657, opened 2018-04-09
feat(@angular-devkit/schematics): allow package name to be specified for install task

mmalerba:


clydin:
commit message should start with `feat(@angular-devkit/schematics):` otherwise this looks ok.

PR #666, opened 2018-04-10
feat(@schematics/angular): save ng new options as defaults

### LINKED ISSUE #615: Regression: no more smart defaults for generate commands in 6.0-rc.0
### Bug Report or Feature Request 
```
```

### Area
```
- [x] devkit
- [x] schematics
```

### Versions
CLI v6.0.0-rc.0

### Repro steps
```
ng new hello --inline-template
ng g component hello
```

Generate a component with an external template instead of an inline one.

### Desired functionality
generate commands should continue as in CLI 1 to use defaults options chosen when creating the app. This feature seems to have completely disappear in CLI 6.0 and it would be a huge step back.

@hansl @Brocco @filipesilva

cyrilletuzi:
Reintroduced #196 in schematics for CLI v6.

Fixes #615

@filipesilva @hansl @Brocco Could you review and merge this, so it it's part of the v6 final release? Thanks.

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { normalize, relative, strings, tags } from '@angular-devkit/core';
+import { JsonObject, normalize, relative, strings, tags } from '@angular-devkit/core';
 import { experimental } from '@angular-devkit/core';
 import {
   MergeStrategy,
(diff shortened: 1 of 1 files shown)


chrste90:
Do you think this is the right location to save the options? I know that it is working to set these options on schematics but shouldn't these be set at application / library scope?

For example the `--prefix` options isn't working too and there is an open PR (https://github.com/angular/devkit/pull/644) which will set the tslint rules to allow different prefixes for libs and apps. I think these defaults here could also be different for apps and libs.

filipesilva:
@Brocco can you have a look?

cyrilletuzi:
Don't think the bazel failure is due to this PR.

About the location of the defaults options, the `ng g library` doesn't contain the same option as `ng new`. So if the defaults are only in the app project options, users would have to redo the config again for each lib. And I think we can assume someone doing a library alongside an app project will use the same defaults. Otherwise, the lib would be a distinct project.

packages/schematics/angular/application/index.ts
```diff
@@ -223,6 +223,36 @@ function addAppToWorkspaceFile(options: ApplicationOptions, workspace: Workspace
     // }
 
     workspace.projects[options.name] = project;
+
+    // tslint:disable-next-line:no-any
+    const schematics: any = {};
+
+    if (options.inlineTemplate !== undefined
```
clydin: none of these should ever be undefined since the schema contains default values for all the options referenced here.

cyrilletuzi: Are you sure? I supposed `options` was the ones passed in the cli command, as the types are `x | undefined`. If so, it's just a quick fix, I will correct the PR later today.

cyrilletuzi: Fixed.

chrste90:
@cyrilletuzi Yes, sounds reasonable :)

I created a separate PR for prefixes: https://github.com/angular/devkit/pull/685

packages/schematics/angular/application/index.ts
```diff
@@ -223,6 +223,36 @@ function addAppToWorkspaceFile(options: ApplicationOptions, workspace: Workspace
     // }
 
     workspace.projects[options.name] = project;
+
+    // tslint:disable-next-line:no-any
```
hansl: Remove `any`, use `JsonObject` instead.

packages/schematics/angular/application/index.ts
```diff
@@ -223,6 +223,36 @@ function addAppToWorkspaceFile(options: ApplicationOptions, workspace: Workspace
...
+    const schematics: any = {};
+
+    if (options.inlineTemplate === true
+    || options.inlineStyle === true
```
hansl: Indent by 4 spaces.

packages/schematics/angular/application/index.ts
cyrilletuzi: Used `any` because it is what is done just above for `project`. Changed to `JsonObject` as request in last commit, but forced me to cast each property.

cyrilletuzi:
@hansl requested changes done :)

PR #667, opened 2018-04-10
@filipesilva fix(@angular-devkit/build-angular): validate fileReplacements, allow string styles/scripts

filipesilva:
Fix angular/angular-cli#5053
Fix angular/angular-cli#10267


packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -67,10 +67,17 @@
           "src": {
```
hansl: Could you `oneOf` this instead? With this current schema we could have `{ replaceWith: "foo" }` which is clearly invalid.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -279,7 +279,7 @@ function extractProjectsConfig(config: CliConfig, tree: Tree): JsonObject {
             ...(isProduction && serviceWorker ? { serviceWorker: true } : {}),
             fileReplacements: [
               {
-                src: `${app.root}/${source}`,
+                source: `${app.root}/${source}`,
```
hansl: The form I'd like to see is `{ replace: "some/file/path", with: "some/other/file" }` which to mean feels more natural. What do you think?

packages/angular_devkit/build_angular/src/browser/schema.json
filipesilva: Yeah that is better.

packages/schematics/angular/migrations/update-6/index.ts
filipesilva: I'm ok with that.

packages/angular_devkit/build_angular/src/browser/schema.d.ts
```diff
@@ -0,0 +1,353 @@
...
+
+  /**
+   * Defines the optimization level of the build.
+   */
+  optimization: boolean;
+
+  /**
+   * Replace files with other files in the build.
+   */
+  fileReplacements: FileReplacements[];
```
larssn: Don't think this type - 'FileReplacements' - made into the source. I've looked around for it, but haven't found it.

filipesilva: @larssn it did make it in, it's something we test for in https://github.com/angular/devkit/blob/master/packages/angular_devkit/build_angular/test/browser/replacements_spec_large.ts

PR #670, opened 2018-04-11
ci: add cli e2e to devkit circle ci

hansl:


alexeagle:
I think we have this in the cli repo?

PR #672, opened 2018-04-11
fix(@schematics/update): allow --from and --to to have partial versions

### LINKED ISSUE #660: Allow any shorthand version for the `--from` argument on @schematics/update:migrate
* `1` should be equivalent to `1.0.0`
* `1.1` should be equivalent to `1.1.0`

hansl:
Basically, --from=1 is the same as --from=1.0.0.

Fix #660

### DIFF
--- a/packages/schematics/update/update/index_spec.ts
+++ b/packages/schematics/update/update/index_spec.ts
@@ -147,6 +147,47 @@ describe('@schematics/update', () => {
             name: 'run-schematic',
             options: jasmine.objectContaining({
               name: 'migrate',
+              options: jasmine.objectContaining({
+                from: '0.1.2',
+                to: '1.6.0',
+              }),
+            }),
+          },
+        ]);
+      }),
+    ).subscribe(undefined, done.fail, done);
+  });
+
+  it('can install and migrate with --from (short version number)', done => {
+    // Add the basic migration package.
+    const content = virtualFs.fileBufferToString(host.sync.read(normalize('/package.json')));
+    const packageJson = JSON.parse(content);
+    packageJson['dependencies']['@angular-devkit-tests/update-migrations'] = '1.6.0';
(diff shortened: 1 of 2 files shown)


packages/schematics/update/update/index_spec.ts
```diff
@@ -153,4 +153,37 @@ describe('@schematics/update', () => {
...
+
+    schematicRunner.runSchematicAsync('update', {
+      packages: ['@angular-devkit-tests/update-migrations'],
+      migrateOnly: true,
+      from: '0',
+    }, appTree).pipe(
+      map(tree => {
+        const packageJson = JSON.parse(tree.readContent('/package.json'));
+        expect(packageJson['dependencies']['@angular-devkit-tests/update-migrations'])
+          .toBe('1.6.0');
```
filipesilva: I don't follow the test. Before running the schematic you set the version, then you run the schematic and check the version hasn't changed.

packages/schematics/update/update/index_spec.ts
```diff
@@ -153,4 +153,37 @@ describe('@schematics/update', () => {
...
+
+        // Check install task.
+        expect(schematicRunner.tasks).toEqual([
+          {
+            name: 'run-schematic',
+            options: jasmine.objectContaining({
+              name: 'migrate',
```
filipesilva: I think this should be checking the install task, like some of the other tests in this file.

packages/schematics/update/update/index_spec.ts
hansl: Yes, it's a migration only update.

packages/schematics/update/update/index_spec.ts
hansl: There's no installation in a migration only.

PR #689, opened 2018-04-12
fix(@angular-devkit/build-angular): strip and use BOM for index.html

clydin:
Fix angular/angular-cli#10210


clydin:
Tests added.  Both UTF-8 with BOM and UTF-16LE with BOM.

PR #690, opened 2018-04-12
fix(@angular-devkit/build-angular): adjust chunk splitting options

clydin:
This ensures that initial vendor modules are alway placed in the vendor bundle.  And also increases the potential for splitting common modules out of async (lazy route) chunks.

Mitigates angular/angular-cli#10294


hansl:
What issue does this PR fixes?  Also, could you provide a _why_ this is needed in the commit message?

clydin:
Updated description.

PR #692, opened 2018-04-12
refactor(@schematics/update): use npm-registry-client to get data

clydin:


packages/schematics/update/update/npm.ts
```diff
@@ -46,23 +44,22 @@ export function getNpmPackageJson(
...
-          subject.error(err);
-        }
-      });
-      response.on('error', err => subject.error(err));
+    const client = new RegistryClient({
+
+    });
+    client.log.level = 'silent';
+    const params = {
+      timeout: 1000,
```
hansl: Timeout needs to be larger.

packages/schematics/update/update/npm.ts
```diff
@@ -46,23 +44,22 @@ export function getNpmPackageJson(
...
-          const json = JSON.parse(data);
-          json.requestedName = packageName;
-          subject.next(json as NpmRepositoryPackageJson);
-          subject.complete();
-        } catch (err) {
-          subject.error(err);
-        }
-      });
-      response.on('error', err => subject.error(err));
+    const client = new RegistryClient({
```
hansl: Are there any proxy information to be passed in?

packages/schematics/update/update/npm.ts
hansl: I'd say 30 seconds is fine. All requests are executed in parallel aynway.

packages/schematics/update/update/npm.ts
clydin: not added yet.  wanted to make sure it was an adequate replacement first.

packages/schematics/update/update/npm.ts
clydin: Done.

PR #695, opened 2018-04-13
fix(@schematics/angular): build in --prod mode by default

cyrilletuzi:
Reintroduce #221, which has (also) disappeared from new schematics...

Would be really nice to include this before final v6. :)

@filipesilva @hansl @Brocco @StephenFluin

### DIFF
--- a/packages/schematics/angular/workspace/files/package.json
+++ b/packages/schematics/angular/workspace/files/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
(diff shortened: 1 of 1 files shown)


hansl:
Please see https://github.com/angular/devkit/pull/629.

TLDR; We will not enable prod builds by default as some people use environment variables that point to production servers (which they should) but don't expect `npm run build` to use prod servers. It's dangerous, and it's not much to change it on your project if that's what you want. Safety first.

Closing this as working as intended.

PR #697, opened 2018-04-13
fix(@angular-devkit/core): Windows path normalize bug

doggy8088:
Fix https://github.com/angular/angular-cli/issues/10328


hansl:
Good catch! Could you add a test for it?  We already have a lot of unittests.

doggy8088:
@hansl Already done! :blush:

Issue #700, opened 2018-04-13
DevServerBuilder: HMR not work


enten:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Area
```
- [x] devkit
- [ ] schematics
```

### Versions

```shell
node --version
v8.9.4
```

```shell
npm --version
5.6.0
```

```shell
npx ng --version

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/


Angular CLI: 6.0.0-rc.4
Node: 8.9.4
OS: linux x64
Angular: 6.0.0-rc.4
... animations, cli, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.5.6
@angular-devkit/build-angular     0.5.6
@angular-devkit/build-optimizer   0.5.6
@angular-devkit/core              0.5.6
@angular-devkit/schematics        0.5.6
@ngtools/json-schema              1.1.0
@ngtools/webpack                  6.0.0-rc.4
@schematics/angular               0.5.6
@schematics/update                0.5.6
rxjs                              6.0.0-ucandoit-rc.6
typescript                        2.7.2
webpack                           4.5.0

```

#### Install @angular/cli@^6.0.0-rc.4 globally
```shell
npm install -g @angular/cli@^6.0.0-rc.4
```

#### Create new project
```shell
ng new example-app
cd example-app
```

#### Start dev server
```shell
ng serve --hmr
```

### The log given by the failure

```
Uncaught Error: [HMR] Hot Module Replacement is disabled.
    at Object../node_modules/webpack/hot/dev-server.js (dev-server.js:60)
    at __webpack_require__ (bootstrap:74)
    at Object.0 (main.ts:12)
    at __webpack_require__ (bootstrap:74)
    at checkDeferredModules (bootstrap:43)
    at Array.webpackJsonpCallback [as push] (bootstrap:30)
    at main.js:1
```

### Desired functionality

[...]

PR #703, opened 2018-04-14
fix(@angular-devkit/build-angular): fix dev server hmr

enten:
Create webpack compiler after webpack config customization.

Fix @angular/devkit#700


filipesilva:
Thanks for finding the problem and fixing it @enten, it will be in the next RC early next week 👍

PR #705, opened 2018-04-14
feat(@schematics/angular): strict option

cyrilletuzi:
--strict option to configure TypeScript in strict mode when generating an app.

Replaces #397

Redo of #462 because of merge issues.

Updated with last schematics, and there is no side effect, so if you want to merge for v6, that would be great.

Note this PR is for v6 only, as `strictFunctionTypes` requires TS 2.6.

@hansl @filipesilva @Brocco

### DIFF
--- a/packages/schematics/angular/ng-new/index_spec.ts
+++ b/packages/schematics/angular/ng-new/index_spec.ts
@@ -46,4 +46,16 @@ describe('Ng New Schematic', () => {
     const content = tree.readContent('/bar/angular.json');
     expect(content).toMatch(/"prefix": "pre"/);
   });
+
+  it('should set strict options in tsconfig.json', () => {
+    const options = { ...defaultOptions, strict: true };
+    const tree = schematicRunner.runSchematic('ng-new', options);
+    const tsconfig = JSON.parse(tree.readContent('/bar/tsconfig.json'));
+
+    expect(tsconfig.compilerOptions.noImplicitAny).toBe(true);
+    expect(tsconfig.compilerOptions.strictNullChecks).toBe(true);
+    expect(tsconfig.compilerOptions.noImplicitThis).toBe(true);
+    expect(tsconfig.compilerOptions.alwaysStrict).toBe(true);
+    expect(tsconfig.compilerOptions.strictFunctionTypes).toBe(true);
+  });
 });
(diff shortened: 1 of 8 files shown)


Brocco (changes requested):
Missing: the logic to pass the option from new to workspace https://github.com/angular/devkit/blob/766059b32e7df49c4537d2dd33256f6ff56abf5b/packages/schematics/angular/ng-new/index.ts#L39

Also, please include tests to verify that the value is set appropriately.

Another note, please also update the documentation to include the option in the CLI repo: https://github.com/angular/angular-cli/blob/master/docs/documentation/new.md

cyrilletuzi:
@Brocco Logic and test added, doc added in angular/angular-cli#10497

packages/schematics/angular/ng-new/index_spec.ts
```diff
@@ -46,4 +46,18 @@ describe('Ng New Schematic', () => {
     const content = tree.readContent('/bar/angular.json');
     expect(content).toMatch(/"prefix": "pre"/);
   });
+
+  it('should set strict options in tsconfig.json', () => {
```
Brocco: This test should live in the workspace schematic, because that is where the logic exists.

Brocco (changes requested):
One comment for the review.

This is a feature and will not be part of 6.0.0, please create an issue associated for this PR so it can be included and tracked for the 6.1.0 release.

packages/schematics/angular/ng-new/index_spec.ts
cyrilletuzi: Are you sure? I put it there on purpose, because even if the logic is in the workspace, the user will use this option from `ng new`, so putting it there also verify that the option is passed well (to avoid problems like in my initial PR where I missed the passing logic). Or maybe do the test twice?

cyrilletuzi:
@hansl @Brocco @StephenFluin Dropped my other features PR, as I've find other ways, but this one would be helpful (as we can't just set `strict: true` as property initialization check does not work well with Angular, so there is currently 5 options to manually change, and there will be more and more in the future).

In the end it's your decision, so let me know if I need to close this PR or not.

alexeagle:
Sorry @cyrilletuzi  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #719, opened 2018-04-17
fix(@angular-devkit/build-optimizer): don't assume enum values

filipesilva:
Enum values were assumed to start with 0 and be digits. But enum values can be anything really.

When a enum declaration was not identified as an enum in a side effect free package, it would be identified as a top level IIFE (in another transform) and assumed pure. This drops the enum.

This PR fixes the behaviour by allowing the enum to be initialized with anything that doesn't have whitespaces. This is still limited and might lead to other false negatives that break things.

We should find a better way of identifying enums.

/cc @clydin

Fix https://github.com/angular/angular/issues/23400


IgorMinar:
can this be merged?

PR #720, opened 2018-04-17
fix(@schematic/angular): e2e should use the prefix

cexbrayat:
Now that the CLI repects the prefix specified, the generated e2e test should also use it.

Currently a project generated with rc.5 fails the e2e test right away.

Another solution is to avoid giving the prefix option to the e2e schematic, and simply revert the title field of `app.component.ts` to `app`. See #721

### DIFF
--- a/packages/schematics/angular/e2e/schema.json
+++ b/packages/schematics/angular/e2e/schema.json
@@ -26,6 +26,13 @@
     "relatedAppName": {
       "description": "The name of the app being tested.",
       "type": "string"
+    },
+    "prefix": {
+      "type": "string",
+      "format": "html-selector",
+      "description": "The prefix to apply to generated selectors.",
+      "default": "app",
+      "alias": "p"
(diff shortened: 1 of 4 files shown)


packages/schematics/angular/e2e/schema.json
```diff
@@ -26,6 +26,10 @@
     "relatedAppName": {
       "description": "The name of the app being tested.",
       "type": "string"
+    },
+    "prefix": {
+      "description": "The prefix to apply.",
+      "type": "string"
```
hansl: Should probably have a default or some way to pickup the value from an existing project on e2e so people can do `ng generate e2e` without a prefix.

cexbrayat: @hansl I added a default (`app` as we do everywhere for prefix. Also added `format: html-selector` and `alias: p` to be coherent.

PR #721, opened 2018-04-17
fix(@schematics/angular): use app as title

cexbrayat:
If the title is dynamic and use the prefix, then the e2e generated test should too.
A simple way to avoid giving the prefix to the e2e schematic is to not use the prefix in the default title.

This is an alternative solution to #720


hansl:
I prefer #720. I'm going to close this one in favor of the other one.

PR #722, opened 2018-04-17
fix(@angular/pwa): correct paths for manifest and icons

### LINKED ISSUE #706: manifest.json comes with incorrect icon paths
```
- [x ] bug report -> please search issues before submitting
- [ ] feature request
```

### Area
```
- [ ] devkit
- [ x] schematics
```

### Versions
Angular CLI: 6.0.0-rc.4

### Repro steps
ng new myApp
cd myApp
ng add @angular/pwa --project=myApp

### Desired functionality
After linking the assets/manifest.json manually (as already mentioned in #686) from the assets folder:
<link rel="manifest" href="./assets/manifest.json">

the referenced icons are failing (404). The manifest is located in the assets folder and so are the manifest icons. However, the icons are referenced from the an images folder, i.e. `src: "images/icons/icon-72x72.png"`

It would be better to locate the manifest in the src folder. This requires:
- move the generated manifest.json file to the src folder
- make a notion in the angular-cli file that manifest.json should be served (similar to the favion.ico)
- change the icon references to the assets folder

cyrilletuzi:
Fixes #706, and the path to `manifest.json` in `index.html` was also wrong.

@hansl @filipesilva @Brocco 

2 side notes:
- beyond my skills, but those errors show tests are missing, because currently `@angular/pwa` was just not working at all, 
- `manifest.json` being in `assets` is risky, it may produce scope problems. `src` would be a better place. It would need custom config like the favicon in `angular.json` (to copy the file when building) and in `ngsw-config.json` (to cache the file).

### DIFF
--- a/packages/schematics/angular/service-worker/index.ts
+++ b/packages/schematics/angular/service-worker/index.ts
@@ -189,7 +189,7 @@ function updateIndexFile(options: ServiceWorkerOptions): Rule {
 
     const indent = getIndent(closingHeadTagLine) + '  ';
     const itemsToAdd = [
-      '<link rel="manifest" href="./manifest.json">',
+      '<link rel="manifest" href="assets/manifest.json">',
       '<meta name="Description" content="Angular Application">',
       '<meta name="theme-color" content="#FFFFFF">',
     ];
--- a/packages/angular/pwa/pwa/files/assets/manifest.json
+++ b/packages/angular/pwa/pwa/files/assets/manifest.json
@@ -4,46 +4,46 @@
   "theme_color": "#1976d2",
   "background_color": "#fafafa",
   "display": "browser",
-  "Scope": "/",
+  "scope": "/",
   "start_url": "/",
   "icons": [
     {
-      "src": "images/icons/icon-72x72.png",
+      "src": "/assets/icons/icon-72x72.png",
       "sizes": "72x72",
       "type": "image/png"
     },
     {
-      "src": "images/icons/icon-96x96.png",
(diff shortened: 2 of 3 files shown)


hansl:
>manifest.json being in assets is risky, it may produce scope problems. src would be a better place. It would need custom config like the favicon in angular.json (to copy the file when building) and in ngsw-config.json (to cache the file).

I verified with @alxhub and it seems that this is fine. "Risky" maybe, but not buggy.

> beyond my skills, but those errors show tests are missing, because currently @angular/pwa was just not working at all,

I'm going to file an issue for that. Our test coverage is basically "generate a schematics, check that files are where we think they should be". We have build tests on the CLI repo, but no run tests for PWA. We need to fix.

hansl:
I'm going to let @brocco do this review.

packages/schematics/angular/service-worker/index.ts
```diff
@@ -189,7 +189,7 @@ function updateIndexFile(options: ServiceWorkerOptions): Rule {
 
     const indent = getIndent(closingHeadTagLine) + '  ';
     const itemsToAdd = [
-      '<link rel="manifest" href="./manifest.json">',
+      '<link rel="manifest" href="assets/manifest.json">',
```
markgoho: I believe the convention is to keep the manifest file in the root of the directory. Is there a reason we're deviating?

markgoho:
This is the first time I've ever seen the manifest _not_ be in the root. Folks are going to look at this schematic as a "gold standard" of sorts for PWAs. I think we should adhere to the pattern that already exists.

PR #723, opened 2018-04-17
fix(@schematics/angular): use string style for new app in karma

cexbrayat:
PR #712 fixed it in a new app, but not in the karma conf.


filipesilva:
Thanks for catching this @cexbrayat 👍

PR #724, opened 2018-04-17
Lazy chunk Webpack config adjustments

clydin:
This prevents initial chunks (most importantly polyfills) from being split.  It also allows the creation of an overall application async common chunk for small common modules if the total is greater than 30Kb.


filipesilva:
This fixes https://github.com/angular/angular-cli/issues/10353.

PR #727, opened 2018-04-17
style: prepend Google @license to all files

theefer:
Prepends the standard @license doc to all `.{js,ts}` files that didn't already have one.


clydin:
Schematic template files shouldn’t have licenses.  They’re part of the user’s application.

theefer:
Ah you're right. I've reverted all files in `packages/schematics`.

PR #733, opened 2018-04-18
feat(@angular-devkit/core): add a JsonSchema visitor that understands refs

hansl:
This is not blocking. I got carried away from fixing 10304.


alexeagle:
@hansl do we still need this? Closing since it's on the devkit repo.

PR #734, opened 2018-04-18
fix(@angular-devkit/build-angular): support some browser options in dev-server

hansl:
This will allow people to override some flags from a base browser target, either from
the command line or from their serve directly.

The options chosen are kinda arbitrary. I simply looked at what made sense.

Fix angular/angular-cli#10304

### DIFF
--- a/packages/angular_devkit/build_angular/src/dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/dev-server/index.ts
@@ -51,6 +51,15 @@ export interface DevServerBuilderOptions {
   watch: boolean;
   hmrWarning: boolean;
   servePathDefaultWarning: boolean;
+
+  optimization?: boolean;
+  aot?: boolean;
+  sourceMap?: boolean;
+  evalSourceMap?: boolean;
+  vendorChunk?: boolean;
+  commonChunk?: boolean;
+  baseHref?: string;
+  progress?: boolean;
 }
 
 interface WebpackDevServerConfigurationOptions {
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/dev-server/schema.json
```diff
@@ -77,6 +77,54 @@
       "type": "boolean",
       "description": "Show a warning when deploy-url/base-href use unsupported serve path values.",
       "default": true
+    },
+    "optimization": {
+      "type": "boolean",
+      "description": "Defines the optimization level of the build.",
+      "default": false
```
filipesilva: These options cannot have a default, otherwise the default will always override the browser target values.

hansl:
> It'd be nice long-term if this type of functionality could be supported without the repetition of options.

Yes, and that's what my other PR that I got carried away with is about; enabling us to walk the schema properly to discover options. Let's not get distracted for 6.0 though and implement `allOf` and `anyOf` and those kind of things in 6.1. This PR is the bandaid we need for 6.0.

packages/angular_devkit/build_angular/src/dev-server/schema.json
hansl: Done.

packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -51,6 +51,15 @@ export interface DevServerBuilderOptions {
   watch: boolean;
   hmrWarning: boolean;
   servePathDefaultWarning: boolean;
+
+  optimization: boolean;
```
filipesilva: These are all optional now.

PR #735, opened 2018-04-18
fix(@schematics/angular): generate app-shell options properly

hansl:
Options are in the "options" key.

Fix angular/angular-cli#10379


hansl:
Just rebased and re-running after the timeout PR.

PR #739, opened 2018-04-18
@angular-devkit: Filter project names

nweldev:
Permit to get only the required project names.

This should permit to add some new behaviours in angular-cli to complete the --project option, for example a new --libraries-only option for the build, lint and test commands.

### DIFF
--- a/packages/angular_devkit/architect/src/architect.ts
+++ b/packages/angular_devkit/architect/src/architect.ts
@@ -78,6 +78,7 @@ export interface BuilderConfiguration<OptionsT = {}> {
 
 export interface TargetSpecifier<OptionsT = {}> {
   project: string;
+  projectFilter?: experimental.workspace.projectFilter;
   target: string;
   configuration?: string;
   overrides?: Partial<OptionsT>;
--- a/packages/angular_devkit/core/src/workspace/workspace.ts
+++ b/packages/angular_devkit/core/src/workspace/workspace.ts
@@ -48,6 +48,8 @@ export class WorkspaceNotYetLoadedException extends BaseException {
 }
 
 
+export type projectFilter = (project: WorkspaceProject, name: string) => boolean;
+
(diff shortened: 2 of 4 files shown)


noelmace:
Is there an issue with AppVeyor ? This PR doesn't impact NgPackagr Build, and test-large is ok on my computer.

clydin:
The test timeouts were recently adjusted.  Try rebasing against master to pick up the changes.

noelmace:
@clydin looks like it's not. Anyway do you have some feedbacks on this PR ?

noelmace:
@filipesilva So, as I said in https://github.com/angular/devkit/issues/730#issuecomment-383220270, this PR is, in my opinion, still relevant even if it's not *the* solution for #730.

In fact, in any mono-repository mixing apps and libs, the CI could be tricky to manage, as we need two separated workflows : one for applications, and one for libraries. When using a unique versioning for a set of libraries, bumping version number, testing, building and releasing should be done for a set of libraries *at once*. It could be for all the libraries in the project, or for a specific subset (for example, all the libraries associated with a given scope (https://github.com/angular/devkit/pull/646 should help a lot on this thanks to @Brocco ).

By adding a filtering solution, we could, instead of just running a bunch of ng commands, run the target on a specific subset of our choice, which should lead to better performances and readability.

Even for applications, we have some cases where we could have specific sets which should have their own CI process separated from the others (for example, a bunch of applications used for one domain).

For now, we have to write some scripts around angular-cli in order to do this kind of thing, what seems to me to be a cheap workaround.

https://github.com/angular/angular-cli/pull/10397 is just a first usage example, which would need some more work.

alexeagle:
Sorry @noelmace - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #746, opened 2018-04-19
feat(@schematics/angular): enable 'no-template-call-expression' rule

kevinbuhmann:
At ng-conf, there was a talk on avoiding binding to function call expression in template to improve change detection performance. Several months ago, I contributed a rule to `codelyzer` that will flag any binding to call expression in template (mgechev/codelyzer#478). This PR enables that rule in the `tslint.json` schematic for Angular projects.

### DIFF
--- a/packages/schematics/angular/workspace/files/tslint.json
+++ b/packages/schematics/angular/workspace/files/tslint.json
@@ -116,6 +116,7 @@
       "check-separator",
       "check-type"
     ],
+    "no-template-call-expression": true,
     "no-output-on-prefix": true,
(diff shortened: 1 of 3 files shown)


kevinphelps:
I just realized that the `no-template-call-expression` rule is not in the published package for `codelyzer`. I have opened mgechev/codelyzer#577 for this.

alexeagle:
Sorry @kevinphelps  - this got lost when we moved back to the angular/angular-cli repo. If you still think this pull request is relevant, could you please re-base on that repo and open a new PR there? Thanks and sorry for the extra churn.

PR #750, opened 2018-04-20
feat(@angular-devkit/build-ng-packagr): permit to set a custom TsConfig file

noelmace:
Add a new tsConfig option in order to permit to use a custom tsConfig file when using @angular-devkit/build-ng-packagr as a build target in angular-cli.

This is, for example, needed in order to resolve this kind of issue : 
```
$ ng build -- --project myLib
BUILD ERROR
projects/common/src/lib/foo/bar.directive.ts(134,41): error TS2339: Property 'includes' does not exist on type 'string[]'.
```
See the evolution of tests for an example. This example show how to change the ES level in lib and target 
All the other options are required for now if we don't want to break the build. I will remove them and bump the ng-packagr dependency version in another PR once / if dherges/ng-packagr#786 is released 

No breaking change.

### DIFF
--- a/packages/angular_devkit/build_ng_packagr/src/build/index.ts
+++ b/packages/angular_devkit/build_ng_packagr/src/build/index.ts
@@ -27,6 +27,7 @@ function requireProjectModule(root: string, moduleName: string) {
 
 export interface NgPackagrBuilderOptions {
   project: string;
+  tsConfig?: string;
 }
 
 export class NgPackagrBuilder implements Builder<NgPackagrBuilderOptions> {
@@ -46,9 +47,15 @@ export class NgPackagrBuilder implements Builder<NgPackagrBuilderOptions> {
         getSystemPath(root), 'ng-packagr') as typeof ngPackagr;
       const packageJsonPath = getSystemPath(resolve(root, normalize(options.project)));
 
-      projectNgPackagr.ngPackagr()
-        .forProject(packageJsonPath)
-        .build()
+      const ngPkgProject = projectNgPackagr.ngPackagr()
+        .forProject(packageJsonPath);
+
+      if (options.tsConfig) {
+        const tsConfigPath = getSystemPath(resolve(root, normalize(options.tsConfig)));
+        ngPkgProject.withTsConfig(tsConfigPath);
+      }
+
+      ngPkgProject.build()
         .then(() => {
           obs.next({ success: true });
           obs.complete();
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ node_modules/
 tmp/
 npm-debug.log*
 yarn-error.log*
+.ng_pkg_build/
 
 # Mac OSX Finder files.
 **/.DS_Store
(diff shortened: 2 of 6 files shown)


clydin:
The flaking test already has a 2 minute timeout.

noelmace:
I pushed the last git "build" commit just in order to test if it works. Will move it to another PR if ok.

noelmace:
@clydin oh, ok, sorry I didn't saw that. Is there something I could do then ? Do you think some of my code break the tests on appveyor for some reason ?

clydin:
It's not your code.  It appears to be a resource issue with AppVeyor.

filipesilva:
Oh this is really nice, I didn't know `ng-packagr` supported a custom tsconfig already!

I'm a bit worried about the mandatory options but that seems like it can go away in the future.

Also wondering if we should also generate a tsconfig when making a library. That would avoid having the linting tsconfig which kinda bothered me as well. WDYT?

noelmace:
@clydin Ok so it's non of my business, I'll just delete this commit then ^^ I was worried that the :x: could make you think that some PRs doesn't deserve review :smile:

packages/angular_devkit/build_ng_packagr/src/build/index.ts
```diff
@@ -46,9 +47,15 @@ export class NgPackagrBuilder implements Builder<NgPackagrBuilderOptions> {
         getSystemPath(root), 'ng-packagr') as typeof ngPackagr;
       const packageJsonPath = getSystemPath(resolve(root, normalize(options.project)));
 
-      projectNgPackagr.ngPackagr()
-        .forProject(packageJsonPath)
-        .build()
+      const ngPkgProject = projectNgPackagr.ngPackagr()
+        .forProject(packageJsonPath);
+
+      if (options.tsConfig) {
```
noelmace: @filipesilva (about your first concern) except if I didn't tought about something, the tsConfig is not mandatory (you can see that the tests doesn't fail after the first commit), given that I first test if it's set before using it in the ngPackagr. Did I forgot some "optionnal" config somewhere ?

noelmace:
@filipesilva (about your second concern) I don't think that adding a tsConfig by default would be a good choice, given that ngPackagr do a lot of custom things with it. For now, there is still some issues with that (see  dherges/ng-packagr#786) so I don't think it's recommanded to use custom tsConfig except you really need it (like in the example in the tests). @dherges ?

FYI : I was already thinking about adding a --tsconfig option to the library schematics with a next PR, but it would realy just be "cherry on the cake", so not a priority

noelmace:
and also @filipesilva "Oh this is really nice" :tada:  thanks a lot :smiley:

filipesilva:
When I mentioned the mandatory options I mean the ones inside the tsconfig (as per the issue you linked), not in the builder here. The option you added is optional and stuff works if we don't add a tsconfig.

So yeah, I think what you added in this PR is ideal for now. Later maybe we can reconsider the tsconfig in a newly generated library.

filipesilva:
Thanks for adding this option!

dherges:
Some additional info / clarifications: therea re several places in ng-packagr where a tsconfig is generated. Summarized:

- a default tsconfig (or base config). this one is customizable
- a tsconfig generated per entry point (the compilation unit that is built). auto-generated, not customizable
- finally, a tsconfig for compilation of the entry point (specific for the compilation target, e.g. es5 and es2015). auto-generated and not customizable.

PR #767, opened 2018-04-24
feat(@angular-devkit/build-angular): add sourcemaps for external libr…

alan-agius4:
At the moment sourcemaps for node_modules are not being loaded.

This is useful for debugging and also if you need this for error reporting during production build, In case you are concerned about build time related performance, we can make this enabled/disabled via a flag.

I'd also suggest to have an option, to set the type of `devtool`, for some users with large apps this will improve the increased the build speed.

PS: I can do the above in a separate Pr, if you like the idea.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9771,6 +9771,42 @@
       "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
       "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w="
     },
+    "source-map-loader": {
+      "version": "0.2.3",
+      "resolved": "https://registry.npmjs.org/source-map-loader/-/source-map-loader-0.2.3.tgz",
+      "integrity": "sha1-1LDIzUfVTtzj5r+g9SP0UrWw5SE=",
+      "requires": {
+        "async": "^2.5.0",
+        "loader-utils": "~0.2.2",
+        "source-map": "~0.6.1"
+      },
+      "dependencies": {
+        "async": {
+          "version": "2.6.1",
+          "resolved": "https://registry.npmjs.org/async/-/async-2.6.1.tgz",
+          "integrity": "sha1-skWiPKcZMAROxT+kaqAKPofGphA=",
+          "requires": {
+            "lodash": "^4.17.10"
+          }
+        },
+        "loader-utils": {
+          "version": "0.2.17",
+          "resolved": "https://registry.npmjs.org/loader-utils/-/loader-utils-0.2.17.tgz",
+          "integrity": "sha1-+G5jdNQyBabmxg6RlvF8Apm/s0g=",
+          "requires": {
+            "big.js": "^3.1.3",
+            "emojis-list": "^2.0.0",
+            "json5": "^0.5.0",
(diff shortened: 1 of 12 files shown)


alan-agius4:
I am a bit worried with all these timeouts

alan-agius4:
What are your thoughts about this @filipesilva?

I also see some other options;
1) make this feature toggable.
2) instead of this feature, add the possibility to add webpack loaders. Shouldn’t be too hard as it can be done from the angular.json since loaders can be configured without being called unlike plugins.
3) continue with this PR as is (just turn it green, still need to figure the increase no. of timeouts)
4) do nothing and don’t support sourcemaps from node_modules.
5) implement more than one of the above.

filipesilva:
@alan-agius4 we orginally removed it because it was very slow and could case warnings for libs that had bad sourcemap paths. Maybe we can re-add it as an option, @hansl @clydin WDYT?

alan-agius4:
Any update @hansl & @clydin?

filipesilva:
We have a couple of questions about this feature:
- what is the performance hit on builds and rebuilds?
- how does this handle cases where the sourcemap ref is faulty or does not exist (this was a problem in the past)?

We could add this as a new flag, but not as changing the current behaviour.

alan-agius4:
Hi @filipesilva , so I gathered some timings which you can see below;

| **Build Type**                                    | **Initial** | **Rebuild** |
|-----------------------------------------------|---------|---------|
| Build Dev - SourceMaps                        | 7456ms  | 150ms   |
| Build Dev - SourceMaps + External SourceMaps  | 16703ms | 135ms   |
| Build Prod - SourceMaps                       | 29379ms | 17886ms |
| Build Prod - SourceMaps + External SourceMaps | 31910ms | 18669ms |

The above are from a standard `ng new` project.

In case a sourcemap is 'broken' a warning will be emitted like the below
```
WARNING in ./node_modules/@angular/core/fesm5/core.js
(Emitted value instead of an instance of Error) Cannot find SourceMap 'core.js.map': Error: Can't resolve './core.js.map' in 'C:\git\ng-cli-sourcemaps-test\node_modules\@angular\core\fesm5'
```

Ps: I deliberately, broke this above sourcemap

alan-agius4:
I have updated the PR, with an additional option. To toggle this feature on/off, I named it `externalSourceMap` however I am open to other suggestions.

filipesilva:
I think the work you did addresses all our concerns. The build speed hit doesn't seem prohibitive, and having an option means it's not mandatory. Having just a warning on broken sourcemaps is good. Your tests and defaults look good to me.

I guess the only question left is the option name. I think `externalSourceMap` is ok, maybe have a slight preference for `vendorSourceMap` but not that strong. @clydin WDYT?

alan-agius4:
@filipesilva, thanks for the review. I actually prefer your suggested naming, `vendorSourceMap`.  I updated the PR, with the new naming. If @clydin has a better naming, I can re-update :)

clydin:
Looks good.  There's a conflict with the package lock.  Are you using npm 6.x btw?  There appears to be a very large diff for the lock file.

alan-agius4:
I am using `npm version 5.5.1`, shall I use v6.x?

PR #768, opened 2018-04-24
Feature Suggestion: Create a HostComponent in all new component specs

happycollision:
This is a feature request (feature change request?) that I had in mind. I did this as a PR so you could easily see what I am suggesting. If you like it, I'll finish off everything and get the PR up to spec.

### Steps Before Complete
- [ ] Get approval of this idea from a member of the angular/devkit team
- [ ] Properly test the changes I've suggested
- [ ] Rewrite commit message to adhere to the proper standards for Angular

I usually end up creating a `HostComponent` for every test I make in Angular. This change extends the current spec setup for a new component to include a bunch of host component boilerplate right from the get-go. I figure that since some components will not use `@Input` or `@Output`, the presence of the host will not alter the test outcomes anyway.

Thoughts are welcome. Especially since there may be a better strategy for testing components which accept `@Input` or produce `@Output` of which I am not aware. I was inspired by the [Angular documentation](https://angular.io/guide/testing#component-inside-a-test-host).

### DIFF
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.component.spec.ts
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.component.spec.ts
@@ -29,7 +29,7 @@ describe('<%= classify(name) %>Component', () => {
   }));
 
   beforeEach(() => {
-    fixture = TestBed.createComponent(<%= classify(name) %>Component);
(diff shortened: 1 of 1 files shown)


alexeagle:
Sorry @happycollision  - this got lost when we moved back to the angular/angular-cli repo. I opened the linked issue there so we will followup sometime.

Issue #775, opened 2018-04-25
Default manifest.json is not well-formed for the app installation + inconsistency + wrong property


webmaxru:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report
- [ ] feature request
```

### Area
```
- [ ] devkit
- [x] schematics
```

### Versions
CLI RC.6

### Repro steps
ng new testApp
ng add @angular/pwa --project testApp
ng build --prod
http-server dist/testApp

Chrome -> DevTools -> Application -> Manifest -> Add to Homescreen

### The log given by the failure
**Errors:**
- Site cannot be installed: one of manifest name or short name must be specified
- Site cannot be installed: the manifest display property must be set to 'standalone' or 'fullscreen'

**Same from Lighthouse audit:**

User will not be prompted to Install the Web App
- Failures: Manifest's `display` value is not one of: minimal-ui | fullscreen | standalone, Manifest does not have `short_name`, Manifest does not have `name`.

Is not configured for a custom splash screen
- Failures: Manifest does not have `name`.

### Desired functionality
The idea to have Web App Manifest is to have the app installable. Proposed fix:
```
{
  "name": "Angular Application",
  "short_name": "Angular App",
  "display": "fullscreen",
  ...
}
```

### Theme-color insonsistency

In index.html:
```
<meta name="theme-color" content="#FFFFFF">
```
should be
```
<meta name="theme-color" content="#1976d2">
```

To follow theme_color from manifest.json as described here: https://developers.google.com/web/fundamentals/web-app-manifest/

> Specify a theme color using the theme_color property. This property sets the color of the toolbar. For this we also suggest duplicating an existing color, specifically the theme-color <meta>.

### Wrong property

```
"splash_pages": null
```
This is not a part of Web App Manifest and should be removed

https://developer.mozilla.org/en-US/docs/Web/Manifest

### References

https://developers.google.com/web/tools/lighthouse/audits/manifest-contains-short_name
Chrome's maximum recommended length is 12 characters.

CC: @Brocco

webmaxru:
I see that `name` and `short_name` should be the title:
https://github.com/angular/devkit/blob/master/packages/angular/pwa/pwa/files/assets/manifest.json#L2

But it doesn't work in CLI RC.6

webmaxru:
Also, Description metatag added to index.html is not about PWA. To be removed

webmaxru:
[...]

PR #776, opened 2018-04-25
fix(@schematics/angular): Make manifest.json correct and consistent

webmaxru:
To fix issues mentioned in https://github.com/angular/devkit/issues/775

1) Removed splash_pages from manifest.json
2) Changed display to fullscreen in manifest.json
2) Updated theme-color in index.html
3) Removed Description metatag from index.html (as it doesn't correlate with PWA)

### DIFF
--- a/packages/angular/pwa/pwa/files/assets/manifest.json
+++ b/packages/angular/pwa/pwa/files/assets/manifest.json
@@ -3,7 +3,7 @@
   "short_name": "<%= title %>",
   "theme_color": "#1976d2",
   "background_color": "#fafafa",
-  "display": "browser",
+  "display": "standalone",
   "scope": "/",
   "start_url": "/",
   "icons": [
@@ -47,6 +47,5 @@
       "sizes": "512x512",
       "type": "image/png"
     }
-  ],
-  "splash_pages": null
+  ]
 }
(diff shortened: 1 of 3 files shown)


packages/angular/pwa/pwa/files/assets/manifest.json
```diff
@@ -3,7 +3,7 @@
   "short_name": "<%= title %>",
   "theme_color": "#1976d2",
   "background_color": "#fafafa",
-  "display": "browser",
+  "display": "fullscreen",
```
Brocco: Please leave the display as browser for the default behavior. Changing it to `fullscreen` could be added as an option for this schematic...
`ng add @angular/pwa --dispaly fullscreen`
But the expected behavior for most users right now should remain as browser.

webmaxru:
@Brocco, with `"display": "browser"` the app cannot be "installed". Requred modes are standalone, fullscreen or minimal-ui (Chrome DevTools output error is a bit wrong because it doesn't list the last one). I believe `minimal-ui` is what we want - a trade-off: app is  installable, but with some UI. Will this work?

https://developer.mozilla.org/en-US/docs/Web/Manifest#display
Btw `minimal-ui`s fallback is exactly `browser`

webmaxru:
Commited with "display": "minimal-ui"

packages/angular/pwa/pwa/files/assets/manifest.json
```diff
@@ -3,7 +3,7 @@
   "short_name": "<%= title %>",
   "theme_color": "#1976d2",
   "background_color": "#fafafa",
-  "display": "browser",
+  "display": "minimal-ui",
```
Brocco: Per conversations, please update the display property to be "standalone"

PR #786, opened 2018-04-26
fix(@schematics/angular): Update environment reference to angular.json

brandonroberts:


packages/schematics/angular/application/files/src/environments/environment.ts
```diff
@@ -1,7 +1,7 @@
 // The file contents for the current environment will overwrite these during build.
 // The build system defaults to the dev environment which uses `environment.ts`, but if you do
 // `ng build --env=prod` then `environment.prod.ts` will be used instead.
```
filipesilva: Can you change the last two lines to the following please? A lot of what's written here is not applicable anymore.
```
// This file can be replaced during build by using the `fileReplacements` array.
// `ng build ---prod` replaces `environment.ts` with `environment.prod.ts`.
// The list of file replacements can be found in `angular.json`.
``

Comment (brandonroberts) [reply packages/schematics/angular/application/files/src/environments/environment.ts]: Sure

Comment (filipesilva) [reply packages/schematics/angular/application/files/src/environments/environment.ts]: Actually, all 4 lines should change to that. Sorry, I was rewording it Midway.

Comment (brandonroberts) [reply packages/schematics/angular/application/files/src/environments/environment.ts]: Ok, let me know if that is correct

File: packages/schematics/angular/application/files/src/environments/environment.ts [hunk 2]
```diff
@@ -1,7 +1,7 @@
-// The file contents for the current environment will overwrite these during build.
-// The build system defaults to the dev environment which uses `environment.ts`, but if you do
-// `ng build --env=prod` then `environment.prod.ts` will be used instead.
-// The list of which env maps to which file can be found in `.angular-cli.json`.
+// The build system defaults to the dev environment which uses `environment.ts`, but
```
filipesilva: The first two lines

```
// The build system defaults to the dev environment which uses `environment.ts`, but
// this file can be replaced during build by using the `fileReplacements` array.
```

Should be replaced with just
```
// This file can be replaced during build by using the `fileReplacements` array.
```

I'm trying to avoid talking about environments, and keep the general ideal to file replacements. This is because the environment doesn't exist at all anymore.

brandonroberts: Fixed

filipesilva:
LGTM, thank you 👍

PR #794, opened 2018-04-26
fix(@schematics/angular): fix lib flatModule options

filipesilva:
These two options not being there breaks AOT builds on consuming apps.

/cc @alan-agius4 @dherges


alan-agius4:
I had a look at this, and I cannot quite understand what was/is the problem as these options are always being set. https://github.com/dherges/ng-packagr/blob/46420ee0069d273338e6217bcbb0250be1efa53a/src/lib/ts/tsconfig.ts#L50

alan-agius4:
I had a look at this, and while they are being set by `ng-packagr` in https://github.com/dherges/ng-packagr/blob/46420ee0069d273338e6217bcbb0250be1efa53a/src/lib/ts/tsconfig.ts#L50

If they are not specified when the tsconfig is parsed ie before hand the `tsconfig.emitFlags` will be `3`, however in order for the metadata to be outputted flattened it should be `7`.

These is one of the `tsconfig` properties, that adding it after the `tsconfig` was changed, has side effects.

PR #795, opened 2018-04-27
fix(@schematics/angular): providedIn in guard

cyrilletuzi:
@hansl @Brocco @filipesilva 

Fixes angular/angular-cli#10509


cyrilletuzi:
AppVeyor error not due to this PR.

Brocco:
We do not want to use `providedIn` for guards as they get directly referenced by routing configuration. This is also why the `module` option was removed from the guard schematic in this PR: #827

PR #796, opened 2018-04-27
fix(@angular-devkit/build-angular): hash files as binary

gkalpak:
For most files it doesn't make a difference, but for files that are not UTF-8 encoded (such as `*.ico`, `*.png`, etc) converting to string before hashing creates a different digest than what the ServiceWorker
will generate (see [here][1]). This in turn causes the SW to think the config is wrong and enter a degraded state.
This commit ensures that `ngsw.json` will contain hashes computed in the same way as the SW will compute them.

[1]: https://github.com/angular/angular/blob/c8a1a14b87e5907458e8e87021e47f9796cb3257/packages/service-worker/worker/src/assets.ts#L418


gkalpak:
Fixed. PTAL

PR #797, opened 2018-04-27
feat(@schematics/angular): make app installable with `@angular/pwa`

gkalpak:
In order for browsers to prompt the user to install a PWA, the `display` property in `manifest.json` must be one of `minimal-ui`, `standalone`, `fullscreen`. ([Source][1])

Partly addresses angular/angular-cli#10236.

[1]: https://developers.google.com/web/tools/lighthouse/audits/install-prompt

### DIFF
--- a/packages/angular/pwa/pwa/files/assets/manifest.json
+++ b/packages/angular/pwa/pwa/files/assets/manifest.json
@@ -3,7 +3,7 @@
   "short_name": "<%= title %>",
   "theme_color": "#1976d2",
   "background_color": "#fafafa",
-  "display": "browser",
+  "display": "minimal-ui",
   "scope": "/",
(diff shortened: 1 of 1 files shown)


packages/angular/pwa/pwa/files/assets/manifest.json
```diff
@@ -3,7 +3,7 @@
   "short_name": "<%= title %>",
   "theme_color": "#1976d2",
   "background_color": "#fafafa",
-  "display": "browser",
+  "display": "minimal-ui",
```
gkalpak: Maybe a different default would be more appropriate. Some boilerplates use `standalone` (and this is also what we use in angular.io), but it depends on the app. `minimal-ui` seems to be the safest.

PR #806, opened 2018-04-27
Allow string assets

filipesilva:


cyrilletuzi:
@filipesilva Is this intended it's now strings when updating but still globs when creating a new project?

filipesilva:
No, that was an oversight. Thank you for pointing that out. Added a fix in https://github.com/angular/devkit/pull/814.

cyrilletuzi:
@filipesilva Updated the schema in angular/angular-cli#10524

PR #813, opened 2018-04-29
fix(@angular/pwa): use default project

cyrilletuzi:
Fixes angular/angular-cli#10469 and additionally fixes angular/angular-cli#10352

@hansl @Brocco  @filipesilva


cyrilletuzi:
Failure in test-large doesn't seem to be related to this PR.

PR #814, opened 2018-04-29
fix(@schematics/angular): use asset strings in new application

filipesilva:


cyrilletuzi:
@filipesilva You missed the test config (aside note: it's quite inconvenient that quite all the config is duplicated for test build).

filipesilva:
@cyrilletuzi good catch, I fixed it now. 

You are right that duplicated config for test is bothersome. Test sits in a weird place in that it uses part of the build pipeline but not the rest of it. So it doesn't really extend the browser config. It's also used to test the libraries for instance.

We didn't come up with a better way to address that so just decided to leave it correct (albeit duplicated) for now.

PR #817, opened 2018-04-30
Fix 10396

hansl:


packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -542,51 +554,90 @@ function updateSpecTsConfig(config: CliConfig): Rule {
...
+      const filesAstNode = findPropertyInAstObject(tsCfgAst, 'files');
+      if (filesAstNode && filesAstNode.kind != 'array') {
+        throw new SchematicsException('Invalid tsconfig "files" property; expected an array.');
       }
+
+      const recorder = host.beginUpdate(tsSpecConfigPath);
+
+      const polyfills = app.polyfills || defaults.polyfills;
+      if (!filesAstNode) {
+        appendPropertyInAstObject(recorder, tsCfgAst, 'files', [polyfills]);
```
filipesilva: If files does not exist, it shouldn't be created. Creating a files array will implicitly exclude all files that are not in the array.

hansl: Done.

PR #835, opened 2018-05-02
update dependencies, fix command for validate-commits, migrates prepush config

ghost:
Hello,

I frequently use the tool `npm-check-updates` to keep my dependencies updated.

I thought it would be a opportunity with the release of 6.0 alongside the recent release of node 10, to update dependencies. node-sass is an important one

I ran the tests and it passed.

I updated everything (except typescript), but I did revert `source-map` and `@types/source-map` due to typings issue.

image


alexeagle:
Sorry @robertbaker we lost this when moving back to the angular/angular-cli repo. Do you think this change is still needed? If so, could you re-base it over there and open a new PR?
Thanks!

PR #838, opened 2018-05-02
fix(@schematics/update): Allow usage strict SSL configuration

Brocco:
Fixes https://github.com/angular/angular-cli/issues/10596

### DIFF
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -91,6 +91,7 @@ export function getNpmPackageJson(
       return concat(
         getNpmConfigOption('proxy'),
         getNpmConfigOption('https-proxy'),
+        getNpmConfigOption('strict-ssl'),
       ).pipe(
         toArray(),
         concatMap(options => {
@@ -100,6 +101,13 @@ export function getNpmPackageJson(
             proxy: {
               http: options[0],
               https: options[1],
+            },
+            ssl: {
+              ...(options[2] === 'false'
+                ? { strict: false }
+                : (options[2] === 'true'
+                   ? { strict: true }
+                   : {})),
             },
           });
           client.log.level = 'silent';


packages/schematics/update/update/npm.ts
```diff
@@ -85,6 +85,7 @@ export function getNpmPackageJson(
       return concat(
         getNpmConfigOption('proxy'),
         getNpmConfigOption('https-proxy'),
+        getNpmConfigOption('strict-ssl'),
```
AlgoTrader: Is getNpmConfigOption('strict-ssl') provides true default when strict-ssl is missing in config?

packages/schematics/update/update/npm.ts
```diff
@@ -95,6 +96,9 @@ export function getNpmPackageJson(
               http: options[0],
               https: options[1],
             },
+            ssl: {
+              strict: options[2],
```
alvaro450: This will give you the string value instead of the boolean. options[2] === 'true' would be more appropriate.

logic01:
Anyone know when we can expect this? I have 5 projects that need upgrading. Or is there a temporary work around? Possibly a list of what this does so I could do it manually?

packages/schematics/update/update/npm.ts
```diff
@@ -101,6 +102,9 @@ export function getNpmPackageJson(
               http: options[0],
               https: options[1],
             },
+            ssl: {
+              strict: options[2] === 'true',
```
clydin: If it is undefined it should stay as undefined to pick up the default.

AlgoTrader: undefined will cause strict: false. strict is default to true. The key should go there only if strict is false

```
if(options[2]==='false') {
  opts = {...opts, {strict: false}}
}
```

ssl stuff should be maximally explicit

clydin: Actually, undefined will cause it to be true.   It is explicitly handled: https://github.com/npm/npm-registry-client/blob/latest/index.js#L32

Brocco: Addressed and updated

PR #851, opened 2018-05-04
fix(@angular-devkit/build-angular): remove mangle and compress from server build

hansl:
Using those for server might break the runtime, and since it all runs on the server
we dont need to compress, local name mangling works fine. Also remove the build
optimizer pass on server optimized builds.

Partial fix for angular/angular-cli#8616 (need to have a fix for 1.7.x)

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -200,6 +200,32 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
     alias = rxPaths(nodeModules);
   } catch (e) { }
 
+  const uglifyOptions = {
+    ecma: wco.supportES2015 ? 6 : 5,
+    warnings: !!buildOptions.verbose,
+    safari10: true,
+    output: {
+      ascii_only: true,
+      comments: false,
+      webkit: true,
+    },
+
+    // On server, we don't want to compress anything.
+    ...(buildOptions.platform == 'server' ? {} : {
+      compress: {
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -200,6 +201,32 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
...
+    warnings: !!buildOptions.verbose,
+    safari10: true,
+    output: {
+      ascii_only: true,
+      comments: false,
+      webkit: true,
+    },
+
+    // On server, we don't want to compress anything.
+    ...(buildOptions.platform == 'server' ? {
```
filipesilva: I think you wanted to do:
```
...({compress: buildOptions.platform == 'server' ? false: {
  //stuff
})
```

hansl: And not false, but `{}`. Done

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -1,6 +1,7 @@
 // tslint:disable
 // TODO: cleanup this file, it's copied as is from Angular CLI.
 
+import { JsonObject } from '@angular-devkit/core';
```
clydin: doesn't appear to be used anywhere

hansl: Done.

PR #853, opened 2018-05-05
feat(@angular/pwa): add content for when javascript is not available

markgoho:
Adding this line to the body will give us a 100 score in lighthouse for PWA.
image
Hosted site (with added `<noscript>` tag): https://angular-cli-pwa.firebaseapp.com/

Thanks so much to @Brocco for providing a great example to follow for manipulating the html file. A few of the updates are to make the variable names more semantic given the similar usage.


christianliebel:
@markgoho Nitpicking: Could you change the casing of `Javascript` to `JavaScript`? That would be excellent. Thank you.

markgoho:
@christianliebel will change later today 👍

PR #858, opened 2018-05-06
fix(@angular-devkit/build-angular): allow verbose in dev-server

mantyz:
This option don't work:
```json
"serve": {
          "options": {
            "verbose": true,
          }
```
 Fixed it.


filipesilva:
At the moment we aren't looking to add more passthrough options to the `dev-server` builder. You can add this option to `browser` builder options and it will used in the `dev-server` as well though.

mantyz:
@filipesilva Thanks for answer. 
I was confused by the fact that you can set this option in `serve.options`, but this do not work.

filipesilva:
That is pretty confusing, and shouldn't be there really. Will remove, thank you for clarifying!

mantyz:
@filipesilva Also remove this option from [angular-cli schema](https://github.com/angular/angular-cli/blob/8e69b3b9b9a3b5a6e87c2eaa9471adcbaa82326b/packages/%40angular/cli/lib/config/schema.json#L1018-L1021), please. Thank you.

PR #862, opened 2018-05-08
fix(@angular-devkit/schematics): add support for BOM to UpdateRecorder

hansl:
Fixes https://github.com/angular/angular-cli/issues/10644

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/recorder.ts
+++ b/packages/angular_devkit/schematics/src/tree/recorder.ts
@@ -21,6 +21,23 @@ export class UpdateRecorderBase implements UpdateRecorder {
     this._path = entry.path;
   }
 
+  static createFromFileEntry(entry: FileEntry): UpdateRecorderBase {
+    const c0 = entry.content.readUInt8(0, true);
+    const c1 = entry.content.readUInt8(1, true);
+    const c2 = entry.content.readUInt8(2, true);
+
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/schematics/src/tree/recorder.ts
```diff
@@ -50,3 +67,22 @@ export class UpdateRecorderBase implements UpdateRecorder {
...
   }
 }
+
+
+export class UpdateRecorderBom extends UpdateRecorderBase {
+  constructor(entry: FileEntry, private _delta = 3) {
+    super(entry);
+  }
+
+  insertLeft(index: number, content: Buffer | string) {
```
clydin: Is `index` in terms of byte offsets or character offsets?  Also is the string for `content` converted into the encoding specified by the BOM?  It might be useful longer term to have both a `BinaryUpdateRecorder` and `TextUpdateRecorder`.

hansl: Indeed. Right now it's in term of binary offset. I'm still unsure how we want to approach this problem.

hansl: We need an `UpdateBuffer` that supports text instead of buffers. That's not a huuuuge challenge, but something we should defer to the libraries (maybe?).

PR #863, opened 2018-05-08
feat(@angular-devkit/build-angular): add watch mode to server

Toxicable:
Adds watch mode into server build

### DIFF
--- a/packages/angular_devkit/build_angular/src/server/index.ts
+++ b/packages/angular_devkit/build_angular/src/server/index.ts
@@ -94,7 +94,14 @@ export class ServerBuilder implements Builder<BuildWebpackServerSchema> {
         };
 
         try {
-          webpackCompiler.run(callback);
+          if (options.watch) {
+            const watching = webpackCompiler.watch({ poll: options.poll }, callback);
+
+            // Teardown logic. Close the watcher when unsubscribed from.
+            return () => watching.close(() => { });
+          } else {
+            webpackCompiler.run(callback);
+          }
         } catch (err) {
           if (err) {
             this.context.logger.error(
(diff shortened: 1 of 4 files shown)


tobi-or-not-tobi:
@Toxicable I got a notion of this PR on your talk at ng-air today - great job! I'm wondering if this is a generic solution, so it can be used for building other apps then universal as well based on file changes. I'm after a watch mode for libraries, see #730. Ideally, we can configure watch modes for specific apps/project.

Toxicable:
@tobi-or-not-tobi No this wont help with your issue sorry.
This is for Universal only, since libraries undergo a much different build/bundle process it cannot be applied to that either.

kondi:
Is it possible to have some explanation or info about "further discussion"? Would be good to have this option soon.

filipesilva (approved):

filipesilva:
@kondi we discussed this last friday and I think this PR is good to get in as is.

mgol:
It doesn't seem to be working correctly for me with `@angular-devkit/build-angular` `0.7.0-beta.1`. Under my project's `architect` key I have the `server` target defined as follows:
```js
"server": {
  "builder": "@angular-devkit/build-angular:server",
  "options": {
    "outputPath": "dist/server",
    "main": "src/main.server.ts",
    "tsConfig": "src/tsconfig.server.json",
    "watch": true
  }
}
```
There is no error when I run:
```
ng run phoenix:server
```
it compiles but it then exits immediately, not entering `watch` mode.

mgol:
Comparing the code of the browser and server parts I see that the browser one doesn't complete the observable in watch mode:
https://github.com/angular/devkit/blob/v6.1.0-beta.1/packages/angular_devkit/build_angular/src/browser/index.ts#L111
while the server one has no special logic for watch here:
https://github.com/angular/devkit/blob/v6.1.0-beta.1/packages/angular_devkit/build_angular/src/server/index.ts#L84
Can that be a problem?

Toxicable:
@mgol Yup it was the problem, just needed a simple check.
Should be fixed in the version that was just released

PR #865, opened 2018-05-08
fix(@schematics/angular): fix e2e schematic to expect welcome message with prefix

fmalcher:
The "application" schematic generates the welcome message in AppComponent with the app prefix.
The "e2e" schematic did not include the prefix in the spec file accordingly.

This PR includes the app prefix in the welcome message spec.
Thanks for reviewing this!


fmalcher:
:-( Duplicate of https://github.com/angular/devkit/pull/720

PR #866, opened 2018-05-08
fix(@angular-devkit/build-angular): fix less loader error

vthinkxie:
close #https://github.com/angular/angular-cli/issues/10430


vthinkxie:
Hi @filipesilva @clydin , anytime to review this pr, it has block all our users to update to angular 6.0 since all components of Ant Design Angular are based on less file.

filipesilva:
@vthinkxie I had a look at https://github.com/angular/angular-cli/issues/10430#issuecomment-390591117 and as far as I can tell the official recommendation (`@plugin` imports) do work right now. If that's not the case let me know in the original issue.

PR #868, opened 2018-05-08
fix(@angular-devkit/build-angular): allow customization of webpack co…

actra-gschuster:
…nfig by providing a hook point / callback that receives the merged config and target name.

In angular.json one may specify "webpackConfig" as path to a JS script that receives the full webpack config object and returns the customized webpack config object.
Those 1% of users that have to modify the config can, the other 99% still can use the simple CLI as intended without ever beeing bugged by this feature.

### DIFF
--- a/packages/angular_devkit/architect/src/architect.ts
+++ b/packages/angular_devkit/architect/src/architect.ts
@@ -72,6 +72,7 @@ export interface BuilderConfiguration<OptionsT = {}> {
   root: Path;
   sourceRoot?: Path;
   projectType: string;
+  target: string;
   builder: string;
   options: OptionsT;
 }
@@ -192,6 +193,7 @@ export class Architect {
       root: project.root as Path,
       sourceRoot: project.sourceRoot as Path | undefined,
       projectType: project.projectType,
+      target: targetName,
       builder: target.builder,
       options: {
         ...options,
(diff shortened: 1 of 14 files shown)


actra-gschuster:
CLA signed

hansl:
Hi @actra-gschuster,

As with the 1.x version of the CLI, we will not allow webpack configuration overrides from the CLI. The good news is we have an actual way forward that might fill your needs, soon.

Cheers.

actra-gschuster:
What's "soon"?
Soon as with I18N?

I don't get it what's your pain-point in allowing experienced developers to customize the config, as explained any un-experienced developer will never ever have to touch the solution provided but those, like me, requiring can do as they please.
Instead of putting off for an unknown time accepting quick fixes would be the way to go - IMHO.

hansl:
> What's "soon"?
We're working on something for 6.1

> Soon as with I18N?
I don't understand the question. i18n is already supported by the CLI.

> I don't get it what's your pain-point in allowing experienced developers to customize the config
It's more complex than "just allow experts to customize stuff". We already have issues filed by people who go to stackoverflow and get told to change their node_modules to customize things.

> Instead of putting off for an unknown time accepting quick fixes would be the way to go - IMHO.
You're not the one who's going to have to maintain those fixes for the next 2 years, we are. We already have an overload of issues for our small (CLI-specific) team, and just opening up the surface area to more bugs won't make anyone happy.

PR #869, opened 2018-05-08
feat: use es2018 lib in tsconfig.json files

christianliebel:
Update the `lib` property in tsconfig.json files of both schematics and the DevKit repo to `es2018`.

The ECMAScript 2018 feature set was finalized in January, so it seems reasonable to upgrade the libs as well. This is specifically interesting for the new Angular libs that are stuck at ECMAScript 2015 (!).

### DIFF
--- a/packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
+++ b/packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
@@ -14,7 +14,7 @@
     "types": [],
     "lib": [
       "dom",
-      "es2015"
+      "es2018"
     ]
   },
   "angularCompilerOptions": {
--- a/packages/schematics/angular/workspace/files/tsconfig.json
+++ b/packages/schematics/angular/workspace/files/tsconfig.json
@@ -13,7 +13,7 @@
       "node_modules/@types"
     ],
     "lib": [
-      "es2017",
+      "es2018",
       "dom"
     ]
   }
--- a/packages/schematics/schematics/blank/project-files/tsconfig.json
+++ b/packages/schematics/schematics/blank/project-files/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "baseUrl": "tsconfig",
     "lib": [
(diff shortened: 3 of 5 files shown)


christianliebel:
I signed it!

clydin:
Please avoid changing files within the `tests` directory.  They are generally setup to test specific scenarios.

Also, `feat` requires a scope.  The commits should ideally be split into:
* `feat(@schematics/angular): ...`
* `feat(@schematics/schematics): ...`
* `build: ...` -> for the main tsconfig

christianliebel:
@clydin I see. Thanks for your help. I’ll update the PR accordingly.

christianliebel:
@clydin The tests looks better now. 👍 However, bazel-build seems to have a problem. Which TypeScript version does bazel use internally? TypeScript 2.7.2 actually has the `es2018` lib.

christianliebel:
I’ve created an additional PR (https://github.com/angular/devkit/pull/870) that only bumps the lib to `es2017`, which should work using the current bazel build.

filipesilva:
@alexeagle do you know why the bazel rules only support TS 2.6.x?

clydin:
The devkit is built against 2.7.  The problem appears to be that the lib type definitions are being sourced from: 
```
/home/circleci/.cache/bazel/_bazel_circleci/9ce5c2144ecf75d11717c0aa41e45a8d/execroot/angular_devkit/external/build_bazel_rules_typescript_tsc_wrapped_deps/node_modules/typescript/lib/lib.es2018.d.ts
```
instead of the typescript version used to compile:
```
/home/circleci/ng/node_modules/typescript/lib/typescript.js
```

clydin:
The error may stem from the following line which attempts to optimize the build time by re-using the library definitions from an existing project.

https://github.com/angular/devkit/blob/master/packages/angular_devkit/build_optimizer/BUILD#L32

alexeagle:
Sorry @christianliebel , this got lost when we moved the code back over to the angular/angular-cli repo. Do you have time to re-base this commit on master over there?

Issue #789, opened 2018-04-26
Restrict usage of some schematics to project type of "application"


Brocco:
- app shell
- pwa
- service worker
- universal

PR #871, opened 2018-05-09
test(@angular-devkit/core): add missing parameter in angular-workspace.json

noelmace:
In a6767dcc, a defaultProject was added to workspaceJson in workspace_spec, which hasn't been
synced with tests/@angular_devkit/workspace/angular-workspace.json as expected. This was breaking
the related tests with more than one project, like in #739.

### DIFF
--- a/tests/@angular_devkit/core/workspace/angular-workspace.json
+++ b/tests/@angular_devkit/core/workspace/angular-workspace.json
@@ -1,6 +1,7 @@
 {
   "version": 1,
   "newProjectRoot": "./projects",
+  "defaultProject": "app",
   "cli": {
     "$globalOverride": "${HOME}/.angular-cli.json",
     "schematics": {


noelmace:
My bad, I missed all the `customWorkspaceJson` behaviours, so I fixed my tests in #789 better following the ones in a6767dc and this doesn't seems to be relevant anymore.

But I still would have a question about this (poke @filipesilva): is the comment about keeping the workspaceJson in sync with the json file still relevant if we can override it when we need ? If so, why ?

Keeping this PR open in order to discuss this and do the according updates.

filipesilva:
@noelmace I don't follow, what comment are you referring to?

noelmace:
@filipesilva Oh sorry for this. This one : https://github.com/angular/devkit/blob/master/packages/angular_devkit/core/src/workspace/workspace_spec.ts#L24

filipesilva:
It should be kept in sync so another test (the one that tests the schema itself) is testing the same thing.

noelmace:
Thanks a lot for your response.

PR #872, opened 2018-05-09
feat(@angular-devkit/core): use the root project as default project by default

noelmace:
If there isn't any defaultProject set, permit to use the "root project" (ie the one which
isn't in the `/projects/` folder) as the default project "by default" if it exists.

In my opinion, if Angular CLI permit to create a project at workspace's root even if their is other projects in the `/projects/` folder (especially as it does this by default when creating a new workspace), this "root project" should be, logically, the default project, as this is the one developers will look at first.

### DIFF
--- a/packages/angular_devkit/core/src/workspace/workspace.ts
+++ b/packages/angular_devkit/core/src/workspace/workspace.ts
@@ -140,6 +140,13 @@ export class Workspace {
     } else if (this.listProjectNames().length === 1) {
       // If there is only one project, return that one.
       return this.listProjectNames()[0];
+    } else {
+      const rootProjectEntry = Object.entries(this._workspace.projects).find(([name, project]) => {
+        return project.root.length === 0;
+      });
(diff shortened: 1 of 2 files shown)


noelmace:
As we are using angular-workspace.json, do you have some advice on how I could / should test this ? In order to avoid creating a conflict with #871, I would need to unset defaultProject in workspaceJson for the related tests, but this would make it unsynced with the json file. Is this a problem ?

filipesilva:
Heya, I'm sorry but I don't think we should assume the default based on root. What you say does make sense, but I think that soft default trades some niche usage convenience (default when multiple projects are present) for some confusion (if things change folder they break). 

If this change were to go in, then it would need to be supported for quite a while. I can see this causing problems in the future and don't think the tradeoff is worth it.

noelmace:
Thank you @filipesilva for this review. Didn't thought about this.

Issue #861, opened 2018-05-07
Multiple entryModule support


artonge:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Area
```
- [x] devkit
- [ ] schematics
```

### Versions

node version: v9.4.0
npm version: 5.6.0
Linux (Ubuntu 17.10)

### Repro steps
Clone this example: https://github.com/artonge/multiple-entrymodules

### The log given by the failure
At execution in the browser I have:
`Uncaught Error: No NgModule metadata found for 'AppModule'.`
This error only appears for the module that wasn't specified has the entryModule in the webpack config.

### Desired functionality
At my company our webpack config has two entry points for two different apps, and generate two bundles for the two apps. The problem is that when using AOT, we can only specify one entryModule, but the two apps have different entryModule. It would be nice to specify two entryModules so the metadata are generated for both of them.

james-criscuolo:
I am also running into this issue while upgrading. I used to be able to use the ngtools/webpack loader without the plugin, which allowed me to get around this issue. I would do this in development, then for production builds I would have the significantly longer build where I sent each entry point individually through webpack. Now I have no choice but to use the plugin in development, so my build process will suffer tremendously once I complete the upgrade.

artonge:
Could this feature be backported to version 1.10.2 ? Any ETA ?
Can a maintainer provide some indications to help me to add this feature ? Where should I look ? How to backport it ?

filipesilva:
This is something we are interested it, but not for 6.x. We will re-evaluate when the Angular Ivy compiler is available as that changes how this feature would work.

sherlock1982:
Could you please provide a kind of a workaround for this? Because I don't quite understand how can I code-split two endpoints in one app. Or for example how a custom element and an app can code-split angular.

artonge:
Thanks for the answer @filipesilva ! As @sherlock1982 asked, could you provide some workaround ? Or tell us if my solution in #875 is safe to use ?

alexeagle:
This issue was moved to angular/angular-cli#12107

PR #873, opened 2018-05-09
feat(@ngtools/webpack): add support for multiple entry modules

artonge:
When the webpack config is building two different apps from two
entry points, one of the apps will not work because metadatas are not
generated. The commit add a new property (entryModules) in
AngularCompilerPluginOptions that permit to regsiter multiple entry
modules.

fix:  #861


artonge:
I didn't add tests as I didn't knew where to put them and how to test correctly. If you give me some hints I can write some.

artonge:
I didn't add tests as I didn't knew where to put them and how to test correctly. If you give me some hints I can write some.

artonge:
Wake up googlebot

PR #875, opened 2018-05-09
feat(@ngtools/webpack): add support for multiple entry modules

### LINKED ISSUE #861: Multiple entryModule support

artonge:
When the webpack config is building two different apps from two
entry points, one of the apps will not work because metadatas are not
generated. The commit add a new property (entryModules) in
AngularCompilerPluginOptions that permit to regsiter multiple entry
modules.

Fix #861

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -67,6 +67,7 @@ export interface AngularCompilerPluginOptions {
   tsConfigPath: string;
   basePath?: string;
   entryModule?: string;
+  entryModules?: string[];
   mainPath?: string;
   skipCodeGeneration?: boolean;
   hostReplacementPaths?: { [path: string]: string };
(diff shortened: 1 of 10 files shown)


artonge:
`npm run test-large` failed for unknown reason. Can a maintainer help sorting this out ?

artonge:
`npm run test-large` failed for unknown reason. Can a maintainer help sorting this out ?

filipesilva:
Please see my response on https://github.com/angular/devkit/issues/861#issuecomment-391797538 as to why we do not want to add this right now.

PR #884, opened 2018-05-10
fix(@schematics/update): respect caret and tilde ranges

filipesilva:
Fix https://github.com/angular/angular-cli/issues/10356

### DIFF
--- a/packages/schematics/update/update/index_spec.ts
+++ b/packages/schematics/update/update/index_spec.ts
@@ -52,6 +52,28 @@ describe('@schematics/update', () => {
     ).subscribe(undefined, done.fail, done);
   }, 45000);
 
+  it('respects existing tilde and caret ranges', done => {
+    // Add ranges.
+    const content = virtualFs.fileBufferToString(host.sync.read(normalize('/package.json')));
+    const packageJson = JSON.parse(content);
+    packageJson['dependencies']['@angular-devkit-tests/update-base'] = '^1.0.0';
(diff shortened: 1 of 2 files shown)


packages/schematics/update/update/index_spec.ts
```diff
@@ -52,6 +52,28 @@ describe('@schematics/update', () => {
...
+    host.sync.write(
+      normalize('/package.json'),
+      virtualFs.stringToFileBuffer(JSON.stringify(packageJson)),
+    );
+
+    schematicRunner.runSchematicAsync('update', { all: true }, appTree).pipe(
+      map(tree => {
+        const packageJson = JSON.parse(tree.readContent('/package.json'));
+        // This one should not change because 1.1.0 was already satisfied by ^1.0.0.
+        expect(packageJson['dependencies']['@angular-devkit-tests/update-base']).toBe('^1.0.0');
```
clydin: wouldn't this cause 1.1.0 to not be used due to an existing lock file?

filipesilva: As far as I can tell, that behaviour is intentional:
```
https://github.com/angular/devkit/blob/master/packages/schematics/update/update/index.ts#L535-L538
```

At this point, the new version is already installed though, and presumably the lockfile is updated.

clydin: OK.  Could envision some issue reports in the future regarding the behavior though.

PR #887, opened 2018-05-10
feat(@ngtools/webpack): resolve all file requests with the virtual fs

### LINKED ISSUE #885: File replacements work only for TypeScript files
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
```

### Area
```
- [X] devkit
- [ ] schematics
```

### Versions
node: v9.9.0
npm: v5.8.0
OS: Fedora 25

### Repro steps
https://github.com/sis0k0/ng-webpack-vfs-issue/

### Desired functionality

The file replacement functionality should work for all types of files.
Currently, the file replacement functionality works only for TS files. More specifically - it works only for files that are part of the webpack compiler host. 
Also, the webpack input filesystem host that's provided as an option of the AngularCompilerPlugin doesn't intercept any requests to html and css files or external  packages, which contain JS (not TS) files.

That behavior is caused by the following method invocations, which turn off the usage of the provided host:

https://github.com/angular/devkit/blob/e335e5daadc66f2814f17f1f15efd9f7984d0e7d/packages/ngtools/webpack/src/virtual_file_system_decorator.ts#L23
https://github.com/angular/devkit/blob/e335e5daadc66f2814f17f1f15efd9f7984d0e7d/packages/ngtools/webpack/src/virtual_file_system_decorator.ts#L31

sis0k0:
fixes #885

### DIFF
--- a/packages/ngtools/webpack/src/virtual_file_system_decorator.ts
+++ b/packages/ngtools/webpack/src/virtual_file_system_decorator.ts
@@ -18,9 +18,8 @@ export class VirtualFileSystemDecorator implements InputFileSystem {
     private _webpackCompilerHost: WebpackCompilerHost,
   ) { }
 
-  // We only need to intercept calls to individual files that are present in WebpackCompilerHost.
   private _readFileSync(path: string): string | null {
-    if (this._webpackCompilerHost.fileExists(path, false)) {
+    if (this._webpackCompilerHost.fileExists(path)) {
       return this._webpackCompilerHost.readFile(path) || null;
     }
 
@@ -28,7 +27,7 @@ export class VirtualFileSystemDecorator implements InputFileSystem {
   }
 
   private _statSync(path: string): Stats | null {
-    if (this._webpackCompilerHost.fileExists(path, false)) {
+    if (this._webpackCompilerHost.fileExists(path)) {
       return this._webpackCompilerHost.stat(path);
     }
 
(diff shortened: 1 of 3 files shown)


packages/ngtools/webpack/src/virtual_file_system_decorator.ts
```diff
@@ -18,17 +18,16 @@ export class VirtualFileSystemDecorator implements InputFileSystem {
     private _webpackCompilerHost: WebpackCompilerHost,
   ) { }
 
-  // We only need to intercept calls to individual files that are present in WebpackCompilerHost.
   private _readFileSync(path: string): string | null {
```
clydin: Webpack filesystems return a Buffer not a string.  This will break any file type that is not text (e.g., JPG, WOFF, etc.)  or uses non-default text encoding (i.e., not UTF8 without a BOM).

sis0k0: I agree. I'll keep digging into that. Do you have some ideas we might try?

clydin: Yes.  Unfortunately, it is a decent amount of work.  The `WebpackCompilerHost` would need to be refactored to internally use `Buffer` objects to store the data and then convert to strings where needed inside the `WebpackCompilerHost`.  And then provide `Buffer` return objects for the read calls within the webpack filesystem decorators.

sis0k0:
Hey @clydin,

I've modified the `VirtualFileStats` class. Now it can be instantiated either with content of type `string`, or content of type `BufferArray`. The other value is generated when needed. Also, some tests are failing/timing out in CircleCI, but they pass locally.

PR #890, opened 2018-05-10
Webpack builder

filipesilva:


filipesilva:
Split into https://github.com/angular/devkit/pull/941, https://github.com/angular/devkit/pull/946 and https://github.com/angular/devkit/pull/947

PR #898, opened 2018-05-11
make ng g universal --clienproject optional: ISSUE #897

ampgular:
It would be good that when the option --clientproject is not available
schematics will pick de defaultproject in angular.json


ampgular:
I thought I knew how to contribute, but I did not,. I will read the docs and re-open agen the pull rquest.
sorry for that.

PR #899, opened 2018-05-11
Validate to prevent additional properties.

Brocco:


packages/schematics/angular/app-shell/index.ts
```diff
@@ -171,12 +170,16 @@ function addUniversalTarget(options: AppShellOptions): Rule {
     // Copy options.
     const universalOptions = {
       ...options,
-      name: options.universalProject,
     };
 
     // Delete non-universal options.
     delete universalOptions.universalProject;
     delete universalOptions.route;
+    delete universalOptions.name;
```
hansl: This is a recurring pattern. Have you thought about adding a `pluck` or `filterKeys` function instead?

Brocco: At this point no, but I think getting this merged is more important than that refactor.

PR #902, opened 2018-05-11
@schematics/angular:universal able to run wo client project options

ampgular:
following the issue #897 I have updated the code in my branch following the contribution guidelines

The goal is when using ng generate universal without the client project option, schematics will take the default project in the angular.json and run. 

As is my first time, if I have done some errors, my apologies.
Thanks


ampgular:
Thanks Mike for taking the time to answer my PR, is awesome!!
I will give a try in the next days.

ampgular:
Hi @Brocco, I am going to close the PR as I did amistake updating the local repository and I lost it.
Nevertheless I will reopen it ith a new number.
I followed your advice and as app-shell is depending in universal must be also changed.
the app-shell is asking two parameters clientproject and universalproject, probably a old ng g universal legacy, as now there is only one project but two different build targets.

PR #905, opened 2018-05-11
CI updates

clydin:


packages/angular_devkit/schematics/src/tree/recorder.ts
```diff
@@ -22,9 +22,9 @@ export class UpdateRecorderBase implements UpdateRecorder {
   }
 
   static createFromFileEntry(entry: FileEntry): UpdateRecorderBase {
-    const c0 = entry.content.readUInt8(0, true);
-    const c1 = entry.content.readUInt8(1, true);
-    const c2 = entry.content.readUInt8(2, true);
+    const c0 = entry.content.length > 0 && entry.content[0];
```
hansl: `readUInt8` already does that check. What are you trying to fix?

hansl: Oh lovely, they removed it in Node 10...

Could you still use `readUInt8` please?  It's clearer, as Buffer might not be an `UInt8Array`

clydin: Done.

PR #906, opened 2018-05-12
feat(@schematics/angular): use application name as title in AppComponent

beeman:
The current schematic creates a new AppComponent with the value of the title property set to `app` (the value of `prefix` in the schematic). 

With the support of multiple projects it makes more sense to use the app name, so if you run `ng generate application admin` the initial page will reflect that: `Welcome to admin`.

### DIFF
--- a/packages/schematics/angular/application/other-files/app.component.spec.ts
+++ b/packages/schematics/angular/application/other-files/app.component.spec.ts
@@ -17,10 +17,10 @@ describe('AppComponent', () => {
     const app = fixture.debugElement.componentInstance;
     expect(app).toBeTruthy();
   }));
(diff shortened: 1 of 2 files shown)


cexbrayat:
Doesn't this break the default e2e test? We just fixed it so it now uses the prefix 😢 https://github.com/angular/devkit/blob/master/packages/schematics/angular/e2e/files/src/app.e2e-spec.ts

beeman:
Good point @cexbrayat - let's get it fixed in the e2e test! Happy to send in a PR for that

cexbrayat:
@beeman Don't bother, it was quick for me to fix it as I already fixed last week :)

@Brocco @hansl See #930

PR #909, opened 2018-05-12
fix(@angular-cli/schematics): set ca from cafile in npm client

rars:
The contents of the cafile can be passed through as ca on the ssl options to the npm client.

This fixes an issue I encountered when running `ng update` with a corporate private repository that requires a cafile in npm config to list trusted certificates, otherwise an error like 'Unable to verify first certificate' is encountered.


hansl:
Hi @rars ,

Could you rebase this on master? We merged some fixes for the `npm.ts` file. Thanks!

rars:
Hi @hansl ,

Sure, I've done that now. Thanks.

PR #911, opened 2018-05-13
feat(@angular-devkit/build-angular): pass config\command line\environment parameters to code

Flybbit:
Add more control on build process:
1) pass parameter known only at build stage to source code using command line argument
   or environment variable
2) define different values in code for different apps and architects
3) strip off some parts of code for some apps and architects

How to use

angular.json:
```json
"defines": [
  {
    "name": "DEFINE_OPTION",
    "value": false
  },
  {
    "name": "ENV_OPTION",
    "value": "process.env.TEST_ENV_DEFINES"
  },
  {
    "name": "COMMAND_LINE_OPTION",
    "value": "customOptions.customCommandLineOption"
  },
  {
    "name": "CALC_OPTION",
    "targets": {
      "production": "283745620252 + 1",
      "development": "2 + 2"
    }
  }
]
```
code:
```ts
declare var DEFINE_OPTION: boolean;
declare var ENV_OPTION: string;
declare var COMMAND_LINE_OPTION: string;
declare var CALC_OPTION: number;
if (DEFINE_OPTION){
  //stripped off
}
```
run:
```shell
set TEST_ENV_DEFINES=env_value
ng build --_customCommandLineOption=42
```

[Issue #10521](https://github.com/angular/angular-cli/issues/10521)


filipesilva:
In https://github.com/angular/angular-cli/issues/10521#issuecomment-390599462 I talked a bit about why we don't want to add this functionality and a current alternative.

PR #920, opened 2018-05-15
docs(aio): fix typo

thebyteman:
Old behavior:
comment on line 2 tells user to use incorrect flag `---prod`

New behavior:
comment on line 2 tells user to use correct flag `--prod`


Brocco (approved):

Brocco:
Please update your commit message, the scope `aio` is not valid and a scope is not needed for `docs`.

You can read more [here](https://github.com/angular/devkit/blob/master/CONTRIBUTING.md#commit).

alexeagle:
Sorry @thebyteman , this got lost when we moved the code back over to the angular/angular-cli repo. Do you have time to re-base this commit on master over there?

PR #924, opened 2018-05-16
Fix circular deps

hansl:


clydin (dismissed):

Brocco (dismissed):

alexeagle:
Looks bad to me - did you take code from tslint-no-circular-imports
https://github.com/bcherny/tslint-no-circular-imports/blob/master/noCircularImportsRule.ts 
and re-license it with Angular MIT?

Why not fork their repo, make the fix there, and point our package.json at your fork?

hansl:
I created https://github.com/bcherny/tslint-no-circular-imports/pull/12

Once the owner makes a release, I'll move to the new version. In the meantime, a lot of our circular dependencies went unnoticed because of the flow.

I'm going to move the tslint rule to use my branch as a dependency in the meantime.

hansl:
Boris merged my PR, I'm doing a rebase right now.

alexeagle:
@hansl this got lost when we moved back to angular-cli repo. Do we have circular deps over there? I guess we have it building with Bazel now so it must have been fixed?

PR #928, opened 2018-05-16
Record host

hansl:


packages/angular_devkit/schematics/src/rules/base.ts
```diff
@@ -106,7 +106,10 @@ export function branchAndMerge(rule: Rule, strategy = MergeStrategy.Default): Ru
     const branchedTree = branch(tree);
 
     return callRule(rule, observableOf(branchedTree), context)
-      .pipe(map(t => staticMerge(tree, t, strategy)));
+      .pipe(
+        last(),
```
clydin: `last` throws if there are no elements.  As long as `callRule` guarantees at least one element, this should be fine.

hansl: All our observables should be containing a value.

PR #930, opened 2018-05-17
fix(@schematics/angular): e2e should use the name

cexbrayat:
PR #906 updated the app schematics to display the name of the app in the title,
so the e2e test has to be fixed.
The prefix is now useless (it was introduced by #720 to fix the e2e test when it was using prefix),
and has been removed.

This a follow up to fix a regression introduced by #906 and released in 6.0.x


filipesilva:
LGTM, thanks for the fix @cexbrayat 👍

PR #936, opened 2018-05-18
Support entry component

JLHwung:
## Summary
Add `entryComponent` support on `@schematics/angular/component`. Fixes angular/angular-cli#7749

## Motivation
We use `@angular/cli` to generate some components and then we use a compiler to imperatively load these components by type, therefore we should add the components to `entryComponents`. While one can copy and paste between `declarations` and `entryComponents`, it will be of convenience to have this switch right in `angular-cli`.

I also do some minor refactors on the `addDeclarationToNgModule` due to duplicated codes.


JLHwung:
Sure, you can squash into one `feat` commit. Personally I am lean to small commit granularity since it is easier to compose into bigger commit.

PR #946, opened 2018-05-21
build-webpack

filipesilva:


packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    const root = this.context.workspace.root;
+
+    const configPath = resolve(root, normalize(builderConfig.options.webpackConfig));
+    const config = this.loadWebpackConfig(getSystemPath(configPath));
+
+    return this.runWebpackDevServer(config);
+  }
+
+  public loadWebpackConfig(webpackConfigPath: string): webpack.Configuration {
+    return require(webpackConfigPath) as webpack.Configuration;
```
clydin: `import` instead of `require`?  would make it promise-based and then could be an observable chain above.

packages/angular_devkit/build_webpack/src/webpack/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        this.context.logger.info(stats.toString(config.stats));
+
+        if (config.watch) {
+          obs.next({ success: !stats.hasErrors() });
+
+          // Never complete on watch mode.
+          return;
+        } else {
+          obs.next({ success: !stats.hasErrors() });
```
clydin: Maybe:
```
obs.next({ success: !stats.hasErrors() });

// Never complete on watch mode.
if (!config.watch) {
  obs.complete();
}
```
Avoids the duplicate `next` call.

packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -353,15 +292,17 @@ export class DevServerBuilder implements Builder<DevServerBuilderOptions> {
 
     config.https = true;
     if (sslKey != null && sslCert != null) {
-      config.key = sslKey;
-      config.cert = sslCert;
+      config.https = {
```
clydin: This seems odd.  `config.https` is being set to `true` above.  Did this work before?

packages/angular_devkit/build_angular/src/extract-i18n/index.ts
```diff
@@ -167,4 +151,14 @@ function getI18nOutfile(format: string) {
   }
 }
 
+export class InMemoryOutputPlugin {
```
clydin: does this need to be exported?

packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
```diff
@@ -68,11 +73,22 @@ export class DevServerBuilder implements Builder<WebpackDevServerBuilderSchema>
...
       );
 
+      // Node 8 has a keepAliveTimeout bug which doesn't respect active connections.
+      // Connections will end after ~5 seconds (arbitrary), often not letting the full download
+      // of large pieces of content, such as a vendor javascript file.  This results in browsers
+      // throwing a "net::ERR_CONTENT_LENGTH_MISMATCH" error.
+      // https://github.com/angular/angular-cli/issues/7197
+      // https://github.com/nodejs/node/issues/13391
+      // https://github.com/nodejs/node/commit/2cb6f2b281eb96a7abe16d58af6ebc9ce23d2e96
+      if (/^v8.\d.\d+$/.test(process.version)) {
```
clydin: We should probably remove this now.  This was corrected in 8.8.  I think we can comfortably tell people to update to the LTS version (8.11) instead of maintaining this.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
filipesilva: I'm not sure the semantics would be the same, nor that it would downlevel to `require`. Do you know?

packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
```diff
@@ -0,0 +1,78 @@
...
+      devServerConfig.port = devServerConfig.port || 8080;
+
+      const statsConfig = devServerConfig.stats || webpackConfig.stats;
+      // Disable stats reporting by the devserver, we have our own logger.
+      devServerConfig.stats = false;
+
+      const webpackCompiler = webpack(webpackConfig);
+      const server = new WebpackDevServer(webpackCompiler, devServerConfig);
+
+      webpackCompiler.hooks.done.tap('build-webpack', (stats: webpack.Stats) => {
```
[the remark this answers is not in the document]
filipesilva: Removed

packages/angular_devkit/build_webpack/src/webpack-dev-server/schema.json
```diff
@@ -0,0 +1,15 @@
+{
+  "title": "Webpack Dev-Server Builder",
+  "description": "Webpack Dev-Server Builder schema for Architect.",
+  "type": "object",
+  "properties": {
+    "webpackConfig": {
```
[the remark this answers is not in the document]
filipesilva: Throughout the other buiders, `fooConfig` is always used. I don't really mind either way, it's just consistent as is. If you feel it should change let me know.

packages/angular_devkit/build_webpack/src/webpack/index.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Path, getSystemPath, normalize, resolve } from '@angular-devkit/core';
+import { Observable } from 'rxjs';
+import * as webpack from 'webpack';
+import { WebpackBuilderSchema } from './schema';
+
+export class BrowserBuilder implements Builder<WebpackBuilderSchema> {
+
+  constructor(public context: BuilderContext) { }
+
+  run(builderConfig: BuilderConfiguration<WebpackBuilderSchema>): Observable<BuildEvent> {
```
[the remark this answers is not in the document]
filipesilva: I kinda feel like making it a base class would end up being more lines to maintain. The only reason why I didn't put the one line `loadWebpackConfig` inside run was because I thought other users would want to load the config from the builder.

Do you have a suggestion that you feel would be better?

packages/angular_devkit/build_webpack/src/webpack/index.ts
filipesilva: Done meanwhile.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -11,21 +11,21 @@ import {
   BuilderConfiguration,
   BuilderContext,
 } from '@angular-devkit/architect';
+import { LoggingCb, WebpackBuilder } from '@angular-devkit/build-webpack';
```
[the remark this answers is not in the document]
filipesilva: Done

packages/angular_devkit/build_angular/src/dev-server/index.ts
filipesilva: It came up when I started using the correct webpack typings. I think it worked, we have a test for it. But also found it weird.

packages/angular_devkit/build_angular/src/extract-i18n/index.ts
filipesilva: No, it doesn't. Removed.

packages/angular_devkit/build_webpack/src/webpack/index.ts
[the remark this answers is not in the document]
clydin: It's fine.  The code count is minimal.  Whenever I see repeat code, I like to question whether there is an opportunity for a refactor.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
filipesilva: SGTM, removed.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
clydin: It does now but won't necessary in the future.  I, personally, prefer the use of `import` for consistency reasons.

packages/angular_devkit/build_webpack/src/webpack-dev-server/schema.json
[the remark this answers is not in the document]
clydin: If the others are already that way then consistency should win.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
filipesilva: Changed to `import`

PR #947, opened 2018-05-21
Eject schematic

filipesilva:
Blocked by https://github.com/angular/devkit/pull/946


alexeagle:
As discussed, we don't plan to bring back the `ng eject` command this way.

PR #954, opened 2018-05-21
refactor: Consolidate adding dependencies

Brocco:


packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,153 @@
...
+import { JsonAstObject, JsonParseMode, parseJsonAst } from '@angular-devkit/core';
+import { SchematicsException, Tree } from '@angular-devkit/schematics';
+import {
+  appendPropertyInAstObject,
+  findPropertyInAstObject,
+} from './json-utils';
+
+
+const pkgJsonPath = '/package.json';
+export enum DependencyType {
```
hansl: `NodeDependencyType`?

packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,153 @@
...
+  Default,
+  Dev,
+  Peer,
+  Optional,
+}
+
+export interface Dependency {
```
hansl: `NodeDependency`

packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,153 @@
...
+    appendPropertyInAstObject(recorder, packageJsonAst, dependencyPropName, {
+      [dependency.packageName]: dependency.version,
+    });
+  } else if (depsNode.kind === 'object') {
+    // check if package already added
+    const depNode = findPropertyInAstObject(depsNode, dependency.packageName);
+
+    if (!depNode) {
+      // No build-angular package, add it.
+      appendPropertyInAstObject(
```
hansl: These need to be sorted. I don't think this is doing it.

packages/schematics/angular/utility/json-utils.ts
```diff
@@ -0,0 +1,114 @@
+/**
```
hansl: Why is still file green and not moved? Migrations is also in the same package; no use in keeping two files with the same code in the same package...

packages/schematics/angular/library/index.ts
```diff
@@ -96,39 +85,103 @@ function updateTsConfig(packageName: string, distRoot: string) {
...
+      buildAngular,
+      ngPackagr,
+      tsickle,
+      tslib,
+      typescript,
+    ]);
 
-    return updateJsonFile(host, 'package.json', (json: PackageJsonPartialType) => {
+    return host;
+    // if (!host.exists('package.json')) { return host; }
```
hansl: Commented code.

hansl (changes requested):
It's a start.

packages/schematics/angular/utility/dependencies.ts
Brocco: renamed

packages/schematics/angular/utility/dependencies.ts
Brocco: renamed

packages/schematics/angular/utility/dependencies.ts
Brocco: The sorting logic is below, anytime a package is added that node is sorted.

packages/schematics/angular/utility/json-utils.ts
Brocco: addressed

packages/schematics/angular/library/index.ts
Brocco: removed

packages/schematics/angular/library/index.ts
```diff
@@ -96,39 +89,70 @@ function updateTsConfig(packageName: string, distRoot: string) {
...
+      packageName: 'tslib',
+      version: '^1.7.1',
+    };
+    const typescript: NodeDependency = {
+      type: NodeDependencyType.Dev,
+      packageName: 'typescript',
+      version: latestVersions.TypeScript,
+    };
+    addPackageJsonDependencies(host, [
+      ngCommonDep,
```
clydin: All these variables are only used once.  An array of object literals would probably be cleaner and less error-prone (it would be easy to forget to add one to the call).

packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,153 @@
...
+export enum NodeDependencyType {
+  Default,
+  Dev,
+  Peer,
+  Optional,
+}
+
+export interface NodeDependency {
+  type: NodeDependencyType;
+  packageName: string;
```
clydin: just `name` would probably be better since it can be a package name, file, GitHub repo, etc.

packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,153 @@
...
+  overwrite?: boolean;
+}
+
+export function addPackageJsonDependencies(tree: Tree, dependencies: NodeDependency[]): void {
+    dependencies
+      .forEach(dependency => addPackageJsonDependency(tree, dependency));
+}
+
+export function addPackageJsonDependency(tree: Tree, dependency: NodeDependency): void {
+  const packageJsonAst = _readPackageJson(tree);
```
clydin: Reparsing the file every time seems inefficient.  Swapping `addPackageJsonDependencies` with `addPackageJsonDependency` as the primary would avoid the repeat parsing.  `addPackageJsonDependency` would then just be a helper that calls the former with a one element array.
An alternative option would be to create a lower level primitive that also takes the parsed JSON AST and then build the two user facing functions on that.

packages/schematics/angular/utility/dependencies.ts
clydin: Instead of using `findPropertyInAstObject`, the function could iterate through the keys and determine both the sorted insertion point and whether it already exists at the same time.

packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  const packageJson = parseJsonAst(content, JsonParseMode.Strict);
+  if (packageJson.kind != 'object') {
+    throw new SchematicsException('Invalid package.json. Was expecting an object');
+  }
+
+  return packageJson;
+}
+
+function _getDependencyPropName(type: NodeDependencyType): string {
```
clydin: Using a string enum for `NodeDependencyType` would allow the elimination of this function.

clydin (changes requested):

packages/schematics/angular/utility/dependencies.ts
clydin: If it is already out of order then this can't really make it any worse.  And if the file is purposefully out of order, should this really change that (non-)order?  Ideally the diff should be minimal either way.

packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,130 @@
...
+    appendPropertyInAstObject(recorder, packageJsonAst, dependency.type, {
+      [dependency.name]: dependency.version,
+    });
+  } else if (depsNode.kind === 'object') {
+    // check if package already added
+    const depNode = findPropertyInAstObject(depsNode, dependency.name);
+
+    if (!depNode) {
+      // Package not found, add it.
+      appendPropertyInAstObject(
```
hansl: We don't want to simply append it. Dependency list in package.json is sorted.

packages/schematics/angular/utility/dependencies.ts
```diff
@@ -0,0 +1,130 @@
...
+    } else if (dependency.overwrite) {
+      // Package found, update version if overwrite.
+      const { end, start } = depNode;
+      recorder.remove(start.offset, end.offset - start.offset);
+      recorder.insertRight(start.offset, JSON.stringify(dependency.version));
+    }
+  }
+
+  tree.commitUpdate(recorder);
+  _sortDependencies(tree, dependency);
```
hansl: Instead of doing this, why not read through the keys and find the first package name that `packageName.localeCompare(dependency.name) > 0` and insert before that key? `_sortDependencies` looks expensive and complex (reading, reparsing and rewriting the file for every key).

hansl: You can call that method `insertPropertyInAstObjectInOrder` and add it to `json-utils`, what do you think?

clydin: suggested that up [here](https://github.com/angular/devkit/pull/954#discussion_r190045801) 😄

Brocco: The issue is that this wouldn't clean up other unsorted packages in the object. Which is the behavior of the npm client.

packages/schematics/angular/utility/json-utils.ts
```diff
@@ -31,6 +37,44 @@ export function appendPropertyInAstObject(
   );
 }
 
+export function removePropertyFromAstObject(
```
hansl: This is unused, and if we were to use it we'd probably want some tests on it. Could you remove it quickly?

Brocco: Removed.

alexeagle:
I think this PR is dead? Please re-open on angular/angular-cli repo otherwise.

PR #956, opened 2018-05-22
test: never use subscribe with done callbacks in jasmine

filipesilva:
Using subscribe with done callbacks in Jasmine leads to `afterEach` being called before the observable is unsubscribed from.

This in turn leads to teardown logic inside observables to only happen after `afterEach` runs.

This can be problematic when the teardown logic in observables interacts with logic in `afterEach`.

For instance, for webpack builders in watch mode, `afterEach` deletes source file folder while the teardown logic closes file watchers. Trying to delete files while they are still being watched leads to errors on Windows.


filipesilva:
Here is an example of the behaviour described above.

Using the common pattern below
```
myObs.pipe(
  // ...
).subscribe(undefined, done.fail, done);
```

and this example
```
describe('rxjs+jasmine behaviour', () => {

  beforeEach(() => console.log('beforeEach'));
  afterEach(() => console.log('afterEach'));

  it('calls afterEach before unsubscribing', (done) => {
    const myObs = new Observable(obs => {
      // Emit one value each second.
      timer(0, 1000).subscribe(val => obs.next(val));

      return () => console.log('myObs teardown logic');
    });

    myObs.pipe(
      tap((nbr) => console.log('emitted', nbr)),
      take(2),
    ).subscribe(
      undefined,
      (err) => {
        console.log('observable emitted error', err.message);
        done.fail();
      },
      () => {
        console.log('observable completed');
        done();
      });
  }, 30000);
});
```
the result will be
```
beforeEach
emitted 0
emitted 1
observable completed
afterEach
myObs teardown logic
```

Notice how `afterEach` is before `myObs teardown logic`. In retrospective, this is obvious enough because `done()` is called inside the subscriptions `complete` callback.

Using the pattern below, we get the desired output:
```
const mySub = myObs.pipe(
  // ...
).subscribe(
  undefined,
  () => {
    mySub.unsubscribe();
    done.fail();
  },
  () => {
    mySub.unsubscribe();
    done();
  },
);
```
```
beforeEach
emitted 0
emitted 1
observable completed
myObs teardown logic
afterEach
````

This is very verbose though. An alternative is to use a promise instead:
```
myObs.pipe(
  tap((nbr) => console.log('emitted', nbr)),
  take(2),
)
.toPromise()
.then(() => {
  console.log('promise resolve');
  done();
}, (err) => {
  console.log('promise error', err.message);
  done.fail();
});
```
```
beforeEach
emitted 0
emitted 1
myObs teardown logic
promise resolved
afterEach
```

Thus I suggest we replace the exiting pattern
```
myObs.pipe(
  // ...
).subscribe(undefined, done.fail, done);
```
with
```
myObs.pipe(
  // ..
).toPromise().then(done, done.fail);
```

This guarantees that logic in `afterEach` blocks always runs after observable teardown logic.

filipesilva:
Maybe it is a good idea to make a lint rule to prevent the old pattern from being used inside spec files.

PR #958, opened 2018-05-22
fix(declaration-files): Fix incorrect declaration files.

mdasberg:
The schema declaration files contained an invalid declaration for:

```
fileReplacements: FileReplacements[];
```
 which should be

```
fileReplacements: FileReplacement[];
```

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/browser/index.ts
@@ -35,7 +35,7 @@ import {
   statsWarningsToString,
 } from '../angular-cli-files/utilities/stats';
 import { addFileReplacements, normalizeAssetPatterns } from '../utils';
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -46,7 +46,7 @@ const webpackMerge = require('webpack-merge');
 // It would really help if it happens during architect.validateBuilderOptions, or similar.
 export interface NormalizedBrowserBuilderSchema extends BrowserBuilderSchema {
   assets: AssetPatternObject[];
-  fileReplacements: CurrentFileReplacement[];
+  fileReplacements: FileReplacement[];
```
filipesilva: This change is not correct. Post normalization the type is `CurrentFileReplacement` instead of `FileReplacement` (which is `DeprecatedFileReplacement | CurrentFileReplacement`).

hansl:
Closing this as the new behaviour would be incorrect.

PR #962, opened 2018-05-23
refactor(@angular-devkit/core): refactor the DevServerBuilder to be e…

mdasberg:
In order to be able to extend Builders, the webpack configuration needs to be overridable.
This pr make it possible to do that. (The same approach as the BrowserBuilder uses.)


mdasberg:
@filipesilva done

PR #965, opened 2018-05-23
feat(@angular-devkit/build-karma): add package

filipesilva:


debug.log
```diff
@@ -0,0 +1 @@
+[0604/171207.114:ERROR:prune_crash_reports.cc(34)] PruneCrashReportDatabase: Failed to get pending reports
```
filipesilva: This file shouldn't be here.

packages/angular_devkit/build_karma/README.md
```diff
@@ -0,0 +1,3 @@
+# Karma Builder for Architect
+
+WIP
```
filipesilva: Needs to be filled in.

filipesilva:
It also looks like the test runner is not exiting. This is likely because of the karma watch tests, since karma doesn't have a good way to exit gracefully when watching.

alexeagle:
Sorry @filipesilva, this got lost when we moved the code back over to the angular/angular-cli repo. Do you have time to re-base this commit on master over there?

PR #970, opened 2018-05-24
Fix JSON parsing

Brocco:
This PR should go in after #954


packages/schematics/package_update/utility/npm.ts
```diff
@@ -93,7 +93,7 @@ function _getNpmPackageJson(
       response.on('data', chunk => data += chunk);
       response.on('end', () => {
         try {
-          const json = JSON.parse(data);
+          const json = parseJson(data, JsonParseMode.Loose);
```
hansl: I don't see why it should be parsed with `Loose` mode. NPM registry is guaranteed to be pure JSON 1.0.

hansl (changes requested):
One comment. If we don't need Loose mode, we shouldn't use it (it's 20x slower than Strict mode). But you can still use the same `jsonParse`, just use Strict mode or the default value. This has the advantage that if an error is caught, it will have the line and character in the message (instead of just the offset).

alexeagle:
Sorry @Brocco , this got lost when we moved the code back over to the angular/angular-cli repo. Do you have time to re-base this commit on master over there?

PR #974, opened 2018-05-24
refactor: Clean up property removal

Brocco:


packages/schematics/angular/utility/clone.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+import { deepCopy } from '@angular-devkit/core';
+
+
+/**
+ * Copy a source with the supplied properties.
+ * @param source Source of the copy.
+ * @param properties Properties to copy.
+ */
+export function copyWith<T>(source: T, properties: string[]): T {
```
hansl: https://stackblitz.com/edit/pick - Here's a version I just wrote that's properly typed.

packages/schematics/angular/utility/clone.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  return copy;
+}
+
+/**
+ * Copy a source without the supplied properties.
+ * @param source Source of the copy.
+ * @param properties Properties to copy.
+ */
+export function copyWithout<T>(source: T, properties: string[]): T {
```
hansl: Removing keys is an anti-pattern. I think we should work from a whitelist instead of a blacklist.

PR #976, opened 2018-05-25
update ng-packagr v3

dherges:


packages/angular_devkit/build_ng_packagr/package.json
```diff
@@ -14,6 +14,6 @@
     "rxjs": "^6.0.0"
   },
   "peerDependencies": {
-    "ng-packagr": "^2.2.0 || ^3.0.0-rc.0"
+    "ng-packagr": "^3.0.0"
```
alan-agius4: Maybe also add the peerDependency to the RC of V4?

alan-agius4:
Tip: Commit scope should be `@angular-devkit/build-ng-packagr`

filipesilva (changes requested):
The changes LGTM, the only problem is the commit message like @alan-agius4 mentioned.

dherges:
Hi @filipesilva, i just updated the PR with the right commit message.

dherges:
A general notice. ng-packagr@3 implements APF 6 (with both ESM's and FESM's) and should work for both Angular 5 and Angular 6.

In ng-packagr@4, there is the change for using [enableResourceInlining in ngc](https://github.com/angular/angular/blob/master/packages/compiler-cli/src/transformers/api.ts#L168-L176) and it thus requires the Angular compiler v6 (I think both `@angular/compiler@6` and `@angular/compiler-cli@6`)

filipesilva:
@dherges SGTM. Nice to hear about the `enableResourceInlining` option! I suppose it lets you remove a fair bit of inlining logic?

filipesilva (dismissed):

filipesilva:
The current CI errors are unrelated to this PR I think:
```
Failures:
1) Schematic Update works with a fixed version
  Message:
    Failed: Unexpected end of JSON input
```

clydin:
Please rebase on master

filipesilva (approved):

packages/angular_devkit/build_ng_packagr/package.json
filipesilva: I was thinking about the peerDeps a bit more, and I think they should be these instead:
```
"peerDependencies": {
     "^2.2.0 || ^3.0.0 || ^4.0.0-rc.0"
}
```

And that the `fix(@angular-devkit/build-ng-packagr): update to ng-packagr v3` commit should be `feat` instead of `fix`.

The reasoning is that changing peerdep support changes how the package behaves, so for users on ng-packagr@2.2.0 they could update build-ng-packagr to the next patch version and get a peerdep warning/error.

This way support for 3.0 and 4 RC is added, support for 2.2 is retained, and users won't get the new build-ng-packagr peerdep errors on a patch.

WDYT?

filipesilva (changes requested):

packages/angular_devkit/build_ng_packagr/package.json
clydin: Removing support for 2.2.0 would be a breaking change.  Adding support for 4.0.0-rc.0 makes sense.  And then with CLI 7.0, support for 2.x (and potentially 3.x) can be dropped.

dherges:
Hi there, I updated the version range for the peerDeps.

filipesilva (approved):
Thanks for the update @dherges!

clydin:
Can you use npm v6 to update the package lock?  There is a large amount of noise in the diff

PR #977, opened 2018-05-28
feat(@angular-devkit/build-ng-packagr): move `test.ts` outside of `src`

alan-agius4:
`test.ts` can ube used by the primary and secondary entrypoints, thus it's best to move it outside of the `src` folder. as With secondary entrypoints one would have the following strcuture;

```
- library
-- src
--- lib (feature)
-- secondary (2nd Entry Point)
--- src
```

See: https://github.com/dherges/ng-packagr/issues/900#issuecomment-392484364


alan-agius4:
Not, quite sure why I am getting
```bash
Failures:
1) Schematic Update works with a fixed version
  Message:
    Failed: Unexpected end of JSON input


```

alexeagle:
Sorry @alan-agius4, this got lost when we moved the code back over to the angular/angular-cli repo. Do you have time to re-base this commit on master over there?

PR #981, opened 2018-05-29
Build Optimizer ES2015 fixes

filipesilva:
These fixes should allow projects using `"target": "es2015",` in `tsconfig.json` to create a working build when using Build Optimizer.

For a new project, using a production build, the main bundle size is:
- ES5: with build optimizer 155 kB, without build optimizer 255 kB (-39%)
- ES2015: with build optimizer 147 kB, without build optimizer 208 kB (-29%)

For Angular.io, using a production build, the main bundle size is:
- ES5: with build optimizer 479 kB, without build optimizer 732 kB (-34%)
- ES2015: with build optimizer 435 kB, without build optimizer 626 kB (-30%)

The improvements are less dramatic but still there.

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
+++ b/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
@@ -37,19 +37,7 @@ const whitelistedAngularModules = [
   /[\\/]node_modules[\\/]@angular[\\/]cdk[\\/]/,
 ];
 
-// TODO: this code is very fragile and should be reworked.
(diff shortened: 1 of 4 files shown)


filipesilva:
AIO doesn't show console errors but doesn't seem to load content properly with these fixes. Investigating.

filipesilva:
The problem in AIO seems to be dropped RxJs code: https://github.com/ReactiveX/rxjs/issues/3756

I am not sure why this is not a problem with Build Optimizer when using the ES5 version of RxJs.

filipesilva:
The issue with AIO was clarified on ReactiveX/rxjs#3756 and is now addressed in this PR.

PR #988, opened 2018-05-30
Transfer CLI to devkit

clydin:
This moves the `@angular/cli` package into this repository.

Note that the E2E tests are not addressed in the PR.


hansl:
Please also add an `ng` binary in `/bin/ng`.

packages/angular/cli/models/architect-command.ts
```diff
@@ -166,34 +166,32 @@ export abstract class ArchitectCommand extends Command<ArchitectCommandOptions>
...
             if (unknownProperty in options) {
               const dashes = unknownProperty.length === 1 ? '-' : '--';
               this.logger.fatal(`Unknown option: '${dashes}${unknownProperty}'`);
-
-              return 1;
             }
           }
           newErrors.push(schemaError);
-        });
+        };
```
hansl: Is that a semicolon at the end of a for loop?

packages/angular/cli/models/architect-command.ts
```diff
@@ -166,34 +166,32 @@ export abstract class ArchitectCommand extends Command<ArchitectCommandOptions>
...
 
         if (newErrors.length > 0) {
           this.logger.error(new schema.SchemaValidationException(newErrors).message);
-
-          return 1;
         }
+
+        return 1;
```
hansl: I'm pretty sure `1` here is the status code, which is error. You mean to do `return 0` here and leave the `return 1` above.

packages/angular/cli/models/architect-command.ts
clydin: yes :).  it's corrected in the lint fix commit.

packages/angular/cli/models/architect-command.ts
clydin: it's inside the catch block.  If it got in there then there's an error.  The return was moved out of above so that all unknown options are reported instead of just the first.

clydin: reverted behavioral change.

PR #989, opened 2018-05-30
fix(@schematics/angular): Transfer baseHref when during update process

Brocco:
fixes angular/angular-cli#10915

### DIFF
--- a/packages/schematics/angular/migrations/update-6/index.ts
+++ b/packages/schematics/angular/migrations/update-6/index.ts
@@ -402,6 +402,7 @@ function extractProjectsConfig(
         index: `${appRoot}/${app.index || defaults.index}`,
         main: `${appRoot}/${app.main || defaults.main}`,
         tsConfig: `${appRoot}/${app.tsconfig || defaults.tsConfig}`,
+        ...(app.baseHref ? { baseHref: app.baseHref } : {}),
         ...buildDefaults,
       };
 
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -431,6 +432,7 @@ function extractProjectsConfig(
       // Serve target
       const serveOptions: JsonObject = {
         browserTarget: `${name}:build`,
+        ...(app.baseHref ? { baseHref: app.baseHref } : {}),
```
filipesilva: It's not needed in the serve, it's just an override for the browser target there. Other than that LGTM.

Brocco: Good catch, updated!

packages/schematics/angular/migrations/update-6/index_spec.ts
```diff
@@ -546,6 +546,24 @@ describe('Migration to v6', () => {
         });
       });
 
+      it('should not set baseHref on build & serve targets if not defined', () => {
+        tree.create(oldConfigPath, JSON.stringify(baseConfig, null, 2));
+        tree = schematicRunner.runSchematic('migration-01', defaultOptions, tree);
+        const build = getConfig(tree).projects.foo.architect.build;
+        expect(build.options.baseHref).toBeUndefined();
+        const serve = getConfig(tree).projects.foo.architect.serve;
+        expect(serve.options.baseHref).toBeUndefined();
```
filipesilva: Nit: doesn't doesn't need to be checked anymore.

PR #990, opened 2018-05-30
fix(@schematics/angular): Fix usage of sourceRoot configuration value

Brocco:
fixes angular/angular-cli#10653

### DIFF
--- a/packages/schematics/angular/class/index.ts
+++ b/packages/schematics/angular/class/index.ts
@@ -22,6 +22,7 @@ import {
 } from '@angular-devkit/schematics';
 import { getWorkspace } from '../utility/config';
 import { parseName } from '../utility/parse-name';
+import { buildDefaultPath } from '../utility/project';
 import { Schema as ClassOptions } from './schema';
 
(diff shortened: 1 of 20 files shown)


packages/schematics/angular/class/index.ts
```diff
@@ -34,7 +34,9 @@ export default function (options: ClassOptions): Rule {
 
     if (options.path === undefined) {
       const projectDirName = project.projectType === 'application' ? 'app' : 'lib';
-      options.path = `/${project.root}/src/${projectDirName}`;
+      options.path = project.sourceRoot
+        ? `/${project.sourceRoot}/${projectDirName}`
+        : `/${project.root}/${project.sourceRoot || 'src'}/${projectDirName}`;
```
clydin: The code section inside this `if` block appears to be repeated in most schematics.  It probably should be pulled out and introduced as a utility function to simplify any changes/fixes in the future.

Brocco: Good idea... refactored.

PR #998, opened 2018-06-01
fix(@angular-devkit/build-optimizer): don't exceed call stack with bi…

filipesilva:
…g files

Related to https://github.com/Microsoft/TypeScript/issues/17033
Fix https://github.com/angular/angular-cli/issues/9636

### DIFF
--- a/packages/angular_devkit/build_optimizer/package.json
+++ b/packages/angular_devkit/build_optimizer/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "loader-utils": "^1.1.0",
     "source-map": "^0.5.6",
-    "typescript": "~2.7.2",
+    "typescript": "~2.9.1",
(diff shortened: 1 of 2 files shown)


filipesilva:
We cannot update the repository itself to use TypeScript 2.9.1 because the Angular Compiler itself does not yet support it.

We can perhaps, update just the `@angular-devkit/build-optimizer` dependency. I manually checked all the build optimizer unit and integration tests and everything seems to pass with TS 2.9.1.

The exception is some of the `prefix-classes` unit tests, where we compare output assuming TS does not preserve all comments (https://github.com/Microsoft/TypeScript/pull/22141). A lot of these seem to be fixed in 2.9.1 so our tests fail there.

Changing the tests to account for comments not being dropped is easy but shouldn't be done right now, because we still test against 2.7.2.

PR #1002, opened 2018-06-01
Transfer legacy e2e

clydin:


packages/angular/cli/models/architect-command.ts
```diff
@@ -267,10 +267,10 @@ export abstract class ArchitectCommand extends Command<ArchitectCommandOptions>
     }
 
     if (!project) {
-      throw new Error('No project specified');
+      project = '';
```
filipesilva: Is this change meant so that Architect itself throws the error instead of throwing it here?

.circleci/config.yml
```diff
@@ -53,6 +53,17 @@ jobs:
...
+  e2e-cli:
+    <<: *defaults
+    environment:
+      BASH_ENV: ~/.profile
+    resource_class: xlarge
+    parallelism: 4
+    steps:
+      - checkout: *post_checkout
+      - restore_cache: *_root_package_lock_key
+      - run: xvfb-run -a npm run test-cli-e2e -- --nb-shards=${CIRCLE_NODE_TOTAL} --shard=${CIRCLE_NODE_INDEX}
```
filipesilva: Nice 👍

tests/legacy-cli/e2e/tests/commands/unknown-option.ts
```diff
@@ -1,9 +1,12 @@
-import { ng } from '../../utils/process';
+import { execAndWaitForOutputToMatch, ng } from '../../utils/process';
+import { expectToFail } from '../../utils/utils';
 
 export default async function() {
-  const { stderr } = await ng('build', '--notanoption');
+  // await expectToFail(() => ng('build', '--notanoption'));
```
filipesilva: Looks like we can drop this comment now.

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -212,22 +214,22 @@ export function useCIChrome(projectDir: string) {
         chromeOptions: {
           args: [
             "--enable-logging",
-            "--no-sandbox",
-            ${process.env['TRAVIS'] ? '"--headless", "--disable-gpu"' : ''}
+            // "--no-sandbox",
```
filipesilva: Looks like we can drop this comment, and the one below, now.

packages/angular/cli/models/architect-command.ts
```diff
@@ -182,8 +182,6 @@ export abstract class ArchitectCommand extends Command<ArchitectCommandOptions>
             if (unknownProperty in options) {
               const dashes = unknownProperty.length === 1 ? '-' : '--';
               this.logger.fatal(`Unknown option: '${dashes}${unknownProperty}'`);
-
-              break;
```
filipesilva: Do these changes supersede https://github.com/angular/angular-cli/pull/10923?

packages/angular/cli/models/architect-command.ts
clydin: The original CLI code didn't have either of the undefined checks (project/target).  They were needed by the stricter TS settings on devkit but apparently the absence of a project or target is actually acceptable at this point in the code.

tests/legacy-cli/e2e/utils/project.ts
clydin: Left them there for now in case appveyor is brought back in.

PR #1015, opened 2018-06-06
fix(@angular-devkit/schematics): include logger in testing context

FrozenPandaz:
SchematicTestRunner.callRule was not using the schematicTestRunner's `logger` requiring the user to pass it in the parent context themselves. This should be default.


FrozenPandaz:
Nevermind. Guess this is more complex than I thought :\ Doesn't seem like `callRule` functions that well right now...

